PR #184, opened 2013-04-17
sanity-check DI configuration

dbu:
i installed dev-master of the mongo bundle in a current symfony standard edition without adding any configuration. the output was 

    [Symfony\Component\DependencyInjection\Exception\InvalidArgumentException]
    Unable to replace alias "doctrine_mongodb.odm.document_manager" with "doctrine.odm.mongodb.document_manager".  

    [Symfony\Component\DependencyInjection\Exception\InvalidArgumentException]
    The service definition "doctrine_mongodb.odm.document_manager" does not exist. 

with this change, we get the more telling message

    [Symfony\Component\DependencyInjection\Exception\InvalidArgumentException]
    You need to define at least one entry in document_managers

### DIFF
--- a/DependencyInjection/DoctrineMongoDBExtension.php
+++ b/DependencyInjection/DoctrineMongoDBExtension.php
@@ -20,6 +20,7 @@ use Symfony\Component\DependencyInjection\Alias;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\DefinitionDecorator;
+use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
 use Symfony\Component\DependencyInjection\Reference;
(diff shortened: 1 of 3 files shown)


DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -39,11 +40,20 @@ public function load(array $configs, ContainerBuilder $container)
...
-        $loader->load('mongodb.xml');
 
         $configuration = new Configuration();
         $config = $this->processConfiguration($configuration, $configs);
 
+        if (! count($config['document_managers'])) {
+            throw new InvalidArgumentException('You need to define at least one entry in doctrine_mongodb.document_managers');
+
+        }
+        if (! count($config['connections'])) {
```
stof: you could use ``->requiresAtLeastOneElement()`` in the config class

dbu:
continued in #293

PR #235, opened 2014-03-25
Update Logger.php

cobaltjeh:
The Symfony LoggerInterface is deprecated, that class hints that the PsrLogger should be used instead.


alcaeus:
Closing this in favor of #323. The upcoming 3.1 version will bump Symfony requirement to 2.3+.

PR #256, opened 2014-06-14
Filter parameters config with expressions

bigfoot90:
This pull is related to https://github.com/doctrine/mongodb-odm/issues/899
It allows set filter parameters in config file.
Parse value parameter by [ExpressionLanguage](http://symfony.com/doc/current/components/expression_language/syntax.html)


instabledesign:
i put that here 
if anyone want something similar
https://github.com/Symftony/MongodbOdmFiltersBundle

PR #269, opened 2014-09-05
Allow passing ObjectManager instance to DocumentType

### LINKED ISSUE #268: Allow option to pass actual DocumentManager to DocumentType
Hi

Is there any particular reason why we can't pass actual ``DocumentManager`` instance to ``DocumentType``? Right now it accepts ``null`` (gets default dm for class from Registry) or ``string`` (gets named dm from Registry). 

Proposed solution: add one more ``if`` in [this normalizer](https://github.com/doctrine/DoctrineMongoDBBundle/blob/master/Form/Type/DocumentType.php#L62)

malarzm:
Fixes #268

### DIFF
--- a/Form/Type/DocumentType.php
+++ b/Form/Type/DocumentType.php
@@ -62,6 +62,10 @@ class DocumentType extends DoctrineType
             if (null === $manager) {
                 return $registry->getManagerForClass($options['class']);
             }
+            
+            if ($manager instanceof ObjectManager) {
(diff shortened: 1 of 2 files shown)


alcaeus:
Merged manually in https://github.com/doctrine/DoctrineMongoDBBundle/commit/bfc117dae240dd5cf0292e265b87fb49194e8051.

PR #289, opened 2015-03-17
Annotation driver mapping pass isn't correctly instantiated

sarcher:
It looks like this was a busted copy/paste from one of the other drivers; all this PR does is correct the arguments so that an annotations-based mappings pass works as expected.

### DIFF
--- a/DependencyInjection/Compiler/DoctrineMongoDBMappingsPass.php
+++ b/DependencyInjection/Compiler/DoctrineMongoDBMappingsPass.php
@@ -128,7 +128,7 @@ class DoctrineMongoDBMappingsPass extends RegisterMappingsPass
      */
     public static function createAnnotationMappingDriver(array $namespaces, array $directories, array $managerParameters, $enabledParameter = false, array $aliasMap = array())
     {
-        $driver = new Definition('Doctrine\ODM\MongoDB\Mapping\Driver\AnnotationDriver', array(new Reference('doctrine_mongodb.odm.metadata.annotation_reader'), $directories));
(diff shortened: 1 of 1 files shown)


DependencyInjection/Compiler/DoctrineMongoDBMappingsPass.php
```diff
@@ -128,9 +128,7 @@ public static function createPhpMappingDriver(array $mappings, array $managerPar
      */
     public static function createAnnotationMappingDriver(array $namespaces, array $directories, array $managerParameters, $enabledParameter = false, array $aliasMap = array())
     {
-        $arguments = array(new Reference('doctrine_mongodb.odm.metadata.annotation_reader'), $directories);
-        $locator = new Definition('Doctrine\Common\Persistence\Mapping\Driver\SymfonyFileLocator', $arguments);
-        $driver = new Definition('Doctrine\ODM\MongoDB\Mapping\Driver\AnnotationDriver', array($locator));
+        $driver = new Definition('Doctrine\ODM\MongoDB\Mapping\Driver\AnnotationDriver', array(new Reference('doctrine_mongodb.odm.metadata.annotation_reader'), $directories));
```
stof: I suggest using ``annotation_driver`` rather than the old alias (which was kept only for BC reasons)

sarcher: I assume you mean `annotation_reader`, right? I made that change -- thanks!

sarcher:
Rebased to current master; let me know if anything else needs to be done in order to get this merged. Thanks!

alcaeus:
Merged manually into 3.0 branch in https://github.com/doctrine/DoctrineMongoDBBundle/commit/bab5072b6dfc855b50c934aa497f0e41051ce1ef. Thank you!

PR #291, opened 2015-04-15
Fix deprecated Factory definition

Soullivaneuh:
This PR replace `setFactoryClass` and `setFactoryMethod` by `setFactory` method, according to Symfony deprecate rules.

`setFactory` was introduce on Symfony 2.6 (https://github.com/symfony/symfony/pull/9839) so this PR will break support of `symfony/dependency-injection` minor to 2.6.

But as you seems to be working on a new major version, this should not be a problem, isn't it?

Regards

### DIFF
--- a/DependencyInjection/DoctrineMongoDBExtension.php
+++ b/DependencyInjection/DoctrineMongoDBExtension.php
@@ -232,8 +232,12 @@ class DoctrineMongoDBExtension extends AbstractDoctrineExtension
             new Reference(sprintf('doctrine_mongodb.odm.%s_connection.event_manager', $connectionName)),
         );
         $odmDmDef = new Definition('%doctrine_mongodb.odm.document_manager.class%', $odmDmArgs);
-        $odmDmDef->setFactoryClass('%doctrine_mongodb.odm.document_manager.class%');
-        $odmDmDef->setFactoryMethod('create');
+        if (method_exists($odmDmDef, 'setFactory')) {
+            $odmDmDef->setFactory(array('%doctrine_mongodb.odm.document_manager.class%', 'create'));
+        } else {
+            $odmDmDef->setFactoryClass('%doctrine_mongodb.odm.document_manager.class%');
+            $odmDmDef->setFactoryMethod('create');
+        }
         $odmDmDef->addTag('doctrine_mongodb.odm.document_manager');
 
         $container
--- a/Tests/DependencyInjection/AbstractMongoDBExtensionTest.php
+++ b/Tests/DependencyInjection/AbstractMongoDBExtensionTest.php
@@ -76,8 +76,12 @@ abstract class AbstractMongoDBExtensionTest extends TestCase
 
         $definition = $container->getDefinition('doctrine_mongodb.odm.default_document_manager');
         $this->assertEquals('%doctrine_mongodb.odm.document_manager.class%', $definition->getClass());
-        $this->assertEquals('%doctrine_mongodb.odm.document_manager.class%', $definition->getFactoryClass());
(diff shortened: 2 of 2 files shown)


stof:
Please write it in a way keeping support for Symfony 2.3+ as 2.3 is a LTS release. See DoctrineBundle for the way to achieve it as @fabpot updated the bundle in a BC way

Soullivaneuh:
@stof Thanks for this advice, it's done. :+1:

stof:
Please update the test assertions too

Soullivaneuh:
Don't have the same result:

```bash
~/projects/fork/DoctrineMongoDBBundle(master)$ phpunit 
PHPUnit 4.4.0 by Sebastian Bergmann.

Configuration read from /home/sullivan/projects/fork/DoctrineMongoDBBundle/phpunit.xml.dist

......S.......................................................... 65 / 78 ( 83%)
.............

Time: 570 ms, Memory: 20.50Mb

OK, but incomplete, skipped, or risky tests!
Tests: 78, Assertions: 343, Skipped: 1.     
```

Soullivaneuh:
Error on test is because of null value on `getFactoryClass` and `getFactoryMethod`.

How I'm supposed to manage test with SF 2.3 support?

Or should I ignoring and change the test only for latest symfony version?

Soullivaneuh:
Maybe run multiple test case according to symfony version? http://docs.travis-ci.com/user/languages/php/#Testing-Against-Multiple-Versions-of-Dependencies-(e.g.-Symfony)

stof:
@Soullivaneuh change the test base on the availability of the method too. See https://github.com/FriendsOfSymfony/FOSUserBundle/pull/1765/files for an example

Soullivaneuh:
Yes I'm currently working on this way.

But I think Travis should be updated too or Symfony <2.6 cases won't be tested anymore. Am I right?

Soullivaneuh:
I tried an update of `.travis.yml`, waiting your opinion.

stof:
@Soullivaneuh I'm on it

.travis.yml
```diff
@@ -7,10 +7,17 @@ php:
 env:
   - MONGO_VERSION=1.2.12
   - MONGO_VERSION=1.3.3
+  - SYMFONY_VERSION=2.1.*
+  - SYMFONY_VERSION=2.2.*
+  - SYMFONY_VERSION=2.3.*
+  - SYMFONY_VERSION=2.4.*
+  - SYMFONY_VERSION=2.5.*
+  - SYMFONY_VERSION=2.6.*
+  - SYMFONY_VERSION=2.7.*
```
stof: Please revert this change:

- it is the wrong way to do this as you make the build matrix explode (especially once adding more PHP versions)
- you are not testing against different Symfony version, only against different FrameworkBundle version (but you are still getting 2.6 for many components, including DependencyInjection)

The right way is to test againts the lowest deps, which is something I'm currently doing and will submit separately

Soullivaneuh:
@stof Revert done.

Soullivaneuh:
Squashed commit for proper PR.

.travis.yml
```diff
@@ -13,4 +13,3 @@ services: mongodb
 before_script:
   - pecl -q install -f mongo-${MONGO_VERSION} && echo "extension=mongo.so" >> `php --ini | grep "Loaded Configuration" | sed -e "s|.*:\s*||"`
   - composer install --dev
-
```
stof: can you revert this useless change ? It will avoid creating conflicts with #292

Soullivaneuh: Missed that. It's done.

stof:
thanks

PR #292, opened 2015-04-15
Improve the Travis setup

stof:
- add testing against newer PHP versions
- add testing on lowest composer deps (and fix them)
- switch to the faster container-based infrastructure
- keep the composer cache between builds
- use the mongo driver shipped by Travis for most builds (it is much newer, and already installed so faster)


jmikola:
Thanks, @stof! :+1:

PR #302, opened 2015-06-18
Allow to load fixtures from files through the command

juchi:
Currently, it's not possible to target specific files when loading the fixtures with the --fixtures option in the command, as it will only take folders into account.
This PR allows to specify files when running the command.


alcaeus:
Merged manually in https://github.com/doctrine/DoctrineMongoDBBundle/commit/71b8fc4ffa955b1378557692e718656c04b8dc5e.

PR #304, opened 2015-07-30
Add PHP 7 on Travis

Soullivaneuh:


jmikola:
The [mongo](http://pecl.php.net/package/mongo) extension (currently required by Doctrine MongoDB) does not compile with PHP7, so there's no practical reason to add PHP7 to the matrix.

Soullivaneuh:
@jmikola Ok but so what next?

Did you plan to change mongo library support?

DoctrineMongoDBBundle will not support PHP7 ever? In that case, which project should we use?

If you plan to support PHP7 later, this PR can be accepted because PHP7 is on `allow_failures` section.

Regards.

stof:
@Soullivaneuh the bundle will support PHP 7 only when mongo itself is supported on PHP 7.

Until this time, making build run and fail on PHP 7 is just wasting Travis resources (we know it has strictly no chance to build as the deps cannot be installed)

Soullivaneuh:
@stof I understand.

In this case, maybe this PR can be still opened and merged when PHP7 will be supported by mongo, as a reminder.

jmikola:
There are no plans to add PHP7 support to the [mongo](https://pecl.php.net/package/mongo) extension, as we're focusing on the newer [mongodb](https://pecl.php.net/package/mongodb) for that. Once we have a stable release of the new extension, I plan to begin work on porting Doctrine and other open-source integrations (e.g. Zend and Symfony's session/cache handlers) over to it.

In a pinch, [mongofill](https://github.com/mongofill/mongofill) could be used to test on PHP7, as folks are doing today for HHVM; however, that's hardly an ideal solution.

PR #306, opened 2015-08-08
Removed incorrect conditional

castarco:
Removed incorrect conditional in findBasePathForBundle. This statement is incompatible with PSR-4 routes, it isn't true that if the str_replace doesn't do any replacement then the route isn't correct.

### DIFF
--- a/Command/DoctrineODMCommand.php
+++ b/Command/DoctrineODMCommand.php
@@ -102,9 +102,10 @@ abstract class DoctrineODMCommand extends ContainerAwareCommand
         $search = str_replace('\\', '/', $bundle->getPath());
         $destination = str_replace('/'.$path, '', $search, $c);
 
-        if ($c != 1) {
-            throw new \RuntimeException(sprintf('Can\'t find base path for bundle (path: "%s", destination: "%s").', $path, $destination));
-        }
+        // TODO: Find a better way to know if we found the base path for the bundle
+        // if ($c != 1) {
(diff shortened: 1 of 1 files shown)


castarco:
As an added comment, I don't want to merge this immediately, I prefer to discuss how to correctly detect "not found paths".

Meanwhile, I'll post here an example of misbehaving with the PSR-4 standard. I have the following namespace, don't blame me for it:

     Zyrcle\Symfony\Bundles\ApiBundle

which the function correctly maps to the following directory:

    /var/www/main_api/src/ZyrcleApiBundle

As you can see, there is no change for str_replace to do a single replacement, so `$c` becomes zero and the conditional throw an exception (in the previous code).

castarco:
OK, I've done some manual tests and it breaks under some conditions, so I'll add more tests to the suite and find another way to go.

castarco:
Hi again, I've found a way to solve the problem :s , but i may break contracts. If the `$c` value becomes 0, then the function could return a 2-tuple (array), with the path and the bundle namespace, this way, the "consumers" of the returned value can infer that there is a mapping between these two strings, and work with the mapping in a more intelligent way.

Currently, the `GenerateRepositoriesDoctrineODMCommand->execute` method is calling the `writeDocumentRepositoryClass` method of a `DocumentRepositoryGenerator` instance. This method is where the problem arises, because it only appends the class "full name" to the path returned by `DoctrineODMCommand::findBasePathForBundle`.

PR #308, opened 2015-08-10
Improved PSR-4 compliance

castarco:
This pull request changes the way how `GenerateRepositoriesDoctrineODMCommand` handles filepaths in order to make this bundle more PSR-4 friendly.

The change is possible because a patch has been merged in the doctrine/mongodb-odm project  https://github.com/doctrine/mongodb-odm/commit/7fe7b97324aee632d8232b6e7db7fe2bbe68512c ).

There's also a minor change (fix) in the `DoctrineODMCommand::findBasePathForBundle` method. The way it was previously done is error prone in systems like Windows. Is easy to view that the "separators normalization" is done in order to make possible the replacement done in the `$destination` variable construction, but this method also works with `DIRECTORY_SEPARATOR` instead than with `'/'`.

An extra comment: I think that there is no need to call `str_replace` in the line *102* (`Command/DoctrineODMCommand.php` file), but I left there to discuss it with other developers.

Probably the pull request needs more tests to be developed before accepting it, I'll posted before typing the tests to allow early discussion.

### DIFF
--- a/Command/DoctrineODMCommand.php
+++ b/Command/DoctrineODMCommand.php
@@ -98,9 +98,9 @@ abstract class DoctrineODMCommand extends ContainerAwareCommand
      */
     protected function findBasePathForBundle($bundle)
     {
-        $path = str_replace('\\', '/', $bundle->getNamespace());
-        $search = str_replace('\\', '/', $bundle->getPath());
-        $destination = str_replace('/'.$path, '', $search, $c);
+        $path = str_replace('\\', DIRECTORY_SEPARATOR, $bundle->getNamespace());
+        $search = str_replace('\\', DIRECTORY_SEPARATOR, $bundle->getPath());
+        $destination = str_replace(DIRECTORY_SEPARATOR.$path, '', $search, $c);
 
         if ($c != 1) {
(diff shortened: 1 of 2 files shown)


malarzm:
As a head up this probably shouldn't be merged until we tag new ODM version

castarco:
@malarzm I'm agree, but I've made a test and I have seen that the last "stable" releases of this bundle are requiring the master branch of the mongodb-odm dependency.

Maybe this should also be fixed to better ensure stability.

malarzm:
@castarco probably yes, but soon we're going stable with ODM so it'll be straighten :)

castarco:
I'm working on the tests, but I think I'll need some advice here.

* Do I have to create a fake bundle to execute the command?
* What's the best approach to mock the `InputInterface` and `OutputInterface` parameters that has the command? I've seen this ( http://www.ricardclau.com/2013/02/testing-symfony2-commands-mocking-the-di-container-with-mockery/ ) , but I don't think it's a good idea to introduce a new dependency on mockery only for this little test.

castarco:
The doctrine/mongodb-odm project is now stable and my changes are available.

alcaeus:
Merged manually in https://github.com/doctrine/DoctrineMongoDBBundle/commit/c257dc75c0a931582b3dc0b4a1a9838158281932. Thanks!

PR #309, opened 2015-08-13
Add a configuration parameter for the fixture loader

nicolasThal:
I have noticed that in the fixture loading command, the loader class is not parametrable.

I have added a configuration parameter to allow the user to use a different loader, as long as it extends the one from the Symfony bridge


nicolasThal:
@alcaeus I have rebased the branch to solve the conflicts.

Thank you for the label :)

alcaeus:
Merged manually in https://github.com/doctrine/DoctrineMongoDBBundle/commit/2fabf17a3962c01b34a1a4b559048e7d9ede81d0. Thanks!

stof:
Defining a service would be better than defining a class name as parameter and assuming the signature of the constructor.

PR #327, opened 2015-12-11
Symfony 3 support

Soullivaneuh:
Cherry-picked @tomasbro commit from #310 for form deprecations.

Code deprecation globally reviewed and corrected.

Please tell me if something is missing.

PR #310, opened 2015-08-17
Fixed use of deprecated method DocumentType::setDefaultOptions().

tomasbro:
As Symfony\Component\Form\FormTypeInterface::setDefaultOptions() method is deprecated in Symfony 2.7, I think it'll be nice to start using \Symfony\Component\Form\AbstractType::configureOptions() instead.

PR #285, opened 2015-02-22
Allow symfony 3.0 components

nicolas-grekas:
Tests should tell if any deprecated interfaces of Symfony are used. If not, then the bundle is defacto compatible with 3.0

PR #328, opened 2015-12-12
Symfony 3.0 support

### LINKED ISSUE #325: Symfony 3.0 support?
DoctrineMongoDBBundle has a dependency on `symfony/framework-bundle: ~2.1`, what means that it's incompatible with Symfony 3.0. When do you guys plan to change it?

alcaeus:
Supersedes #327, #310 and #285
Closes #325 

Thank you @tomasbro and @Soullivaneuh for getting started on this. I have taken your work for 3.0 compatibility and put it in here.

With the coming version (either 3.1 or 4.0, which is TBD) we are removing support for all Symfony versions older than 2.8 and in turn introduce compatibility with Symfony 3.0. We will tag one last release of the 3.0 line to have a 2.x version out that requires a stable version of doctrine/mongodb-odm. Please note that we are not providing any long term support for versions of our bundle that work with LTS versions of Symfony.

While there are a few open issues I'd like to look at in the coming days, getting this merged will allow people to use dev-master should they be using Symfony 3.0. I'm still hoping to get a release out rather sooner than later, but there are a few issues that fell through the cracks that I think are worth at least looking at.

One reason of concern is our test coverage, I know that the form types are poorly tested. Any PRs in this direction which can help timely adoption of future Symfony versions are greatly appreciated.

### DIFF
--- a/Form/Type/DocumentType.php
+++ b/Form/Type/DocumentType.php
@@ -18,7 +18,7 @@ use Doctrine\Bundle\MongoDBBundle\Form\ChoiceList\MongoDBQueryBuilderLoader;
 use Doctrine\Common\Persistence\ObjectManager;
 use Symfony\Bridge\Doctrine\Form\Type\DoctrineType;
 use Symfony\Component\OptionsResolver\Options;
-use Symfony\Component\OptionsResolver\OptionsResolverInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
  * Form type for a MongoDB document
@@ -41,11 +41,11 @@ class DocumentType extends DoctrineType
     }
 
     /**
-     * @param OptionsResolverInterface $resolver
+     * @param OptionsResolver $resolver
      */
-    public function setDefaultOptions(OptionsResolverInterface $resolver)
+    public function configureOptions(OptionsResolver $resolver)
     {
-        parent::setDefaultOptions($resolver);
+        parent::configureOptions($resolver);
 
         $resolver->setDefaults(array(
             'document_manager' => null,
@@ -66,16 +66,6 @@ class DocumentType extends DoctrineType
             return $registry->getManager($manager);
         };
 
-        $resolver->setNormalizers(array(
-            'em' => $normalizer,
-        ));
-    }
-
-    /**
-     * @see Symfony\Component\Form\FormTypeInterface::getName()
-     */
-    public function getName()
-    {
-        return 'document';
+        $resolver->setNormalizer('em', $normalizer);
     }
 }
--- a/Tests/Form/Type/DocumentTypeTest.php
+++ b/Tests/Form/Type/DocumentTypeTest.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\Bundle\MongoDBBundle\Tests\Form\Type;
 
+use Doctrine\Bundle\MongoDBBundle\Form\Type\DocumentType;
 use Doctrine\Bundle\MongoDBBundle\Tests\TestCase;
 use Doctrine\Bundle\MongoDBBundle\Form\DoctrineMongoDBExtension;
 use Symfony\Component\Form\Test\TypeTestCase;
@@ -30,7 +31,7 @@ class DocumentTypeTest extends TypeTestCase
 
     public function testDocumentManagerOptionSetsEmOption()
     {
-        $field = $this->factory->createNamed('name', 'document', null, array(
+        $field = $this->factory->createNamed('name', DocumentType::CLASS, null, array(
             'class' => 'Doctrine\Bundle\MongoDBBundle\Tests\Fixtures\Form\Document',
             'document_manager' => 'default',
         ));
@@ -39,11 +40,11 @@ class DocumentTypeTest extends TypeTestCase
     }
 
     /**
-     * @expectedException InvalidArgumentException
+     * @expectedException \InvalidArgumentException
      */
     public function testSettingDocumentManagerAndEmOptionShouldThrowException()
     {
-        $field = $this->factory->createNamed('name', 'document', null, array(
+        $field = $this->factory->createNamed('name', DocumentType::CLASS, null, array(
             'document_manager' => 'default',
             'em' => 'default',
         ));
--- a/composer.json
+++ b/composer.json
@@ -12,19 +12,20 @@
     ],
     "require": {
         "php": "^5.5",
-        "doctrine/mongodb-odm": "~1.0",
-        "psr/log": "~1.0",
-        "symfony/options-resolver": "~2.1",
-        "symfony/doctrine-bridge": "~2.1",
-        "symfony/framework-bundle": "~2.1",
-        "symfony/dependency-injection": "~2.2",
-        "symfony/console": "~2.1",
-        "symfony/config": "~2.2"
+        "doctrine/mongodb-odm": "^1.0",
+        "psr/log": "^1.0",
+        "symfony/options-resolver": "^2.8 || ^3.0",
+        "symfony/doctrine-bridge": "^2.8 || ^3.0",
+        "symfony/framework-bundle": "^2.8 || ^3.0",
+        "symfony/dependency-injection": "^2.8 || ^3.0",
+        "symfony/console": "^2.8 || ^3.0",
+        "symfony/config": "^2.8 || ^3.0"
     },
     "require-dev": {
         "doctrine/data-fixtures": "@dev",
-        "symfony/form": "~2.3",
-        "symfony/yaml": "~2.1"
+        "symfony/form": "^2.8 || ^3.0",
+        "symfony/yaml": "^2.8 || ^3.0",
+        "symfony/phpunit-bridge": "^2.8 || ^3.0"
     },
     "suggest": {
         "doctrine/data-fixtures": "Load data fixtures"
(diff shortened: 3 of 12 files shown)


zemd:
:thumbsup:

BertrandD:
:thumbsup:

peshi:
It would be very nice to see this in dev-master asap. :+1:

Soullivaneuh:
Looks good for the global part.

IMHO, as I said in #327 I would like to see my fixes on a new patch version which support both ~2.3 and ~3.0 Symfony versions.

This is actually working. I have some difficulties to understand why this could not happen.

alcaeus:
The decision was made to drop support for old Symfony versions, partly to avoid adding that ```setDefaultOptions```/```configureOptions``` compatibility code. That's the reason why this is targeting master.

Form/Type/DocumentType.php
```diff
@@ -66,16 +66,22 @@ public function setDefaultOptions(OptionsResolverInterface $resolver)
             return $registry->getManager($manager);
         };
 
-        $resolver->setNormalizers(array(
-            'em' => $normalizer,
-        ));
+        $resolver->setNormalizer('em', $normalizer);
```
xabbuh: You then have to bump the required version of the OptionsResolver component to `2.6` or higher.

alcaeus: https://github.com/doctrine/DoctrineMongoDBBundle/pull/328/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R17 ;)

Form/Type/DocumentType.php
```diff
@@ -66,16 +66,22 @@ public function setDefaultOptions(OptionsResolverInterface $resolver)
...
     }
 
     /**
-     * @see Symfony\Component\Form\FormTypeInterface::getName()
+     * {@inheritdoc}
      */
     public function getName()
     {
+        return $this->getBlockPrefix();
```
xabbuh: Why don't you rely on the FQCN here?

Tests/Form/Type/DocumentTypeTest.php
```diff
@@ -30,7 +30,10 @@ public function setUp()
 
     public function testDocumentManagerOptionSetsEmOption()
     {
-        $field = $this->factory->createNamed('name', 'document', null, array(
+        $typeClass = method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')
+            ? 'Doctrine\Bundle\MongoDBBundle\Form\Type\DocumentType' : 'document';
```
xabbuh: This is not needed as you already require the Form component in 2.8 or higher.

alcaeus: Right - so starting with 2.8 forms are created using FQCN? Excuse the silly questions, I still haven't gotten past 2.6 ;)

xabbuh: Exactly :)

xabbuh: So basically you can just rely on the FQCN here and remove `setName()` and `setBlockPrefix()` from the form type.

Form/Type/DocumentType.php
alcaeus: Reading through the [UPGRADE document for 2.8](https://github.com/symfony/symfony/blob/2.8/UPGRADE-2.8.md) we have to return what is now the block prefix here for BC reasons.

xabbuh: This method is only here for BC with Symfony versions before 2.8. Starting with 3.0 it's completely unused by the core.

alcaeus: I could swear I saw it in the FormTypeInterface just a minute ago. You're right again. Thanks!

Form/Type/DocumentType.php
```diff
@@ -66,15 +66,13 @@ public function setDefaultOptions(OptionsResolverInterface $resolver)
...
     }
 
     /**
-     * @see Symfony\Component\Form\FormTypeInterface::getName()
+     * {@inheritdoc}
      */
-    public function getName()
+    public function getBlockPrefix()
     {
         return 'document';
```
xabbuh: Overriding this method shouldn't be necessary at all.

alcaeus: Yes, wanted to do that along with getName but forgot.

xabbuh:
The code looks good to me. Maybe you want to update the Travis configuration to run jobs with both Symfony 2.8 and 3.0.

Soullivaneuh:
@alcaeus Ok now I understand, but I disagree. :-)

IMHO, the code tricks to keep SF 2.3 LTS support are not to painful and I thinks this should be kept as long as SF 2.3 will be supported.

After all, the decision is up to you. ;-)

Soullivaneuh:
> The decision was made to drop support for old Symfony versions, partly to avoid adding that setDefaultOptions/configureOptions compatibility code. That's the reason why this is targeting master.

:+1:

Tests/Form/Type/DocumentTypeTest.php
```diff
@@ -30,7 +30,7 @@ public function setUp()
 
     public function testDocumentManagerOptionSetsEmOption()
     {
-        $field = $this->factory->createNamed('name', 'document', null, array(
+        $field = $this->factory->createNamed('name', 'Doctrine\Bundle\MongoDBBundle\Form\Type\DocumentType', null, array(
```
Soullivaneuh: If you support PHP 5.5+, you should use `DocumentType::class` and import the statement.

Tests/Form/Type/DocumentTypeTest.php
```diff
@@ -43,7 +43,7 @@ public function testDocumentManagerOptionSetsEmOption()
      */
     public function testSettingDocumentManagerAndEmOptionShouldThrowException()
     {
-        $field = $this->factory->createNamed('name', 'document', null, array(
+        $field = $this->factory->createNamed('name', 'Doctrine\Bundle\MongoDBBundle\Form\Type\DocumentType', null, array(
```
Soullivaneuh: If you support PHP 5.5+, you should use `DocumentType::class` and import the statement.

composer.json
```diff
@@ -11,20 +11,21 @@
...
+        "php": "^5.5",
         "doctrine/mongodb-odm": "~1.0",
         "psr/log": "~1.0",
-        "symfony/options-resolver": "~2.1",
-        "symfony/doctrine-bridge": "~2.1",
-        "symfony/framework-bundle": "~2.1",
-        "symfony/dependency-injection": "~2.2",
-        "symfony/console": "~2.1",
-        "symfony/config": "~2.2"
+        "symfony/options-resolver": "~2.8|~3.0",
```
Soullivaneuh: Maybe it's time to follow Composer new conventions.

`~2.8` -> `^2.8`.

`^2.8|^3.0` -> `^2.8 || ^3.0`

Tests/Form/Type/DocumentTypeTest.php
alcaeus: Good idea!

composer.json
alcaeus: Sometimes I keep forgetting about these. Changed, thanks!

alcaeus:
@Soullivaneuh 
> IMHO, the code tricks to keep SF 2.3 LTS support are not to painful and I thinks this should be kept as long as SF 2.3 will be supported.

That is exactly why I decided to leave these out: we're not doing LTS releases. I'll just quote @Ocramius here ([Source](https://github.com/doctrine/cache/pull/109#issuecomment-162181547)):
> We do security support and we support what we can, but we can't nor will commit to having LTS stuff anyway.

We're busy enough keeping everything together in terms of MongoDB releases, the new driver, PHP 7 support for ODM and so on. Keeping an old version supported just to fulfil the LTS promise of another package is far down on my priority list. Hint: we're not going to support 2.8 until its EOL (11/2019) either.

alcaeus:
@xabbuh
> Maybe you want to update the Travis configuration to run jobs with both Symfony 2.8 and 3.0.

The Travis build is already testing that. by using ```--prefer-lowest``` in the build matrix.

stof:
@alcaeus in your case, this is a Symfony bundle, so taking care of the Symfony maintenance is quite important. Your target users are users of the framework.
Dropping support for Symfony LTS releases before the end of maintenance would mean that you would end up with 2 choices:
- providing LTS releases of the bundle before the bump, so that people can keep using the Symfony LTS while still being compatible with uptodate MongoDB driver
- telling people that want to use LTs versions of Symfony that using MongoDB is not an option for them

IMO, none of these options is good. the first option gives you **more** work than keeping compat with Symfony LTS releases (as you have to maintain 2 codebases), while the second option simply cannot happen (people will still want to use MongoDB)
The right option IMO is to keep support for Symfony LTS versions as long as possible in your bundle, allowing you to maintain a single version of the bundle at a time (using BC layers for Symfony 2.7 is much less work than maintaining 2 versions). See MonologBundle for instance. All users (even people using Symfony 2.3 LTS) are using MonologBundle 2.8.2 (assuming they haven't locked an older version of course, but they should ``composer update`` then), precisely because we don't do LTS releases of the bundle.

Soullivaneuh:
I totally agree with @stof comment.

alcaeus:
@stof We'll be providing bugfixes for 2.3 and 2.7 in the current 3.0 branch.

peshi:
@alcaeus Thank you for merging this - and a wise decision indeed.

Soullivaneuh:
Any ETA for 3.1 version push?

malarzm:
We'll try this weekend

PR #338, opened 2016-01-20
Deprecate mapping drivers provided by bundle

### LINKED ISSUE #311: YAML / XML Mapping Driver
Are those still needed?

https://github.com/doctrine/DoctrineMongoDBBundle/tree/master/Mapping/Driver

see:
https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Mapping/Driver/SimplifiedXmlDriver.php
https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Mapping/Driver/SimplifiedYamlDriver.php

alcaeus:
Closes #311.


stof:
Please fix tests

alcaeus:
@stof Tests are now failing due to tests for the deprecated classes. Should I remove those tests or change them to use the ODM drivers and test against those?

alcaeus:
Ping @stof: suggestions? Otherwise I'll just remove tests since they cause build to fail due to deprecation warnings.

PR #340, opened 2016-01-22
Fix DoctrineMongoDBTypeGuesser

ChubV:
Fix incompatibility of guessed type names with latest form component.


malarzm:
@ChubV thanks for the PR! If I may I'll hold this one open for a moment as I'm trying to evaluate how much work will be needed to have Symfony 2.7 support

ChubV:
As you wish. I can use a patch for a while. But it breaks FOSUserBundle registration form now, for example (and it is why i made this PR).

malarzm:
@ChubV ok, indeed, I'm not sure if I'll be able to work on this today... could you just add new lines at the end of files where they're missed? I'll merge afterwards

malarzm:
Merged manually in https://github.com/doctrine/DoctrineMongoDBBundle/commit/48900a076c03691a1c3f1e32292fa89400fcd6d2 - thanks @ChubV and sorry for the delay, we don't get notifications about commits and I haven't checked earlier :)

PR #341, opened 2016-01-23
Improve continuous integration

alcaeus:
This PR adds phpunit as a dev dependency to composer.json. It also adds code coverage generation to the travis build and uploads the coverage to scrutinizer-ci, which would have to be enabled for this project. This would allow us to track code coverage and start working on adding more tests for the bundle.


alcaeus:
Superseded by #381.

PR #342, opened 2016-01-31
Symfony 2.7 compatibility

malarzm:


malarzm:
Ok, I went through Symfony's 2.8 changelog and it seems these changes should be enough... @stof @alcaeus do you recall anything else that needs to be worked around?

alcaeus:
@malarzm Can't think of any - IIRC the type stuff was the only thing changed in 2.8.

Form/DoctrineMongoDBTypeGuesser.php
```diff
@@ -39,9 +40,12 @@ class DoctrineMongoDBTypeGuesser implements FormTypeGuesserInterface
 
     private $cache = array();
 
+    private $typeFQCN;
+
     public function __construct(ManagerRegistry $registry)
     {
         $this->registry = $registry;
+        $this->typeFQCN = version_compare(Kernel::VERSION, '2.8.0', '>=');
```
stof: this should not detect the kernel version but the version of the Form component. It is perfectly valid to install the Form 2.8 component with HttpKernel 2.7.

Use ``method_exists('\Symfony\Component\Form\AbstractType', 'getBlockPrefix')`` to detect the new form component

stof:
IIRC, the decision to require 2.8+ was made to simplify the need to handle BC layers (and 2.8 is the current LTS)

Form/DoctrineMongoDBTypeGuesser.php
malarzm: will fix later, thanks for the tip :+1:

malarzm:
Personally if we'll get away with only these changes I'll be fine with supporting 2.7

PR #344, opened 2016-02-07
doc: fix yaml syntax

mhor:
see [https://github.com/symfony/symfony-standard/issues/426](https://github.com/symfony/symfony-standard/issues/426)

### DIFF
--- a/Resources/doc/config.rst
+++ b/Resources/doc/config.rst
@@ -39,7 +39,7 @@ Sample Configuration
         doctrine_mongodb:
             connections:
                 default:
-                    server: %mongodb_server%
+                    server: "%mongodb_server%"
 
 If you wish to use memcache to cache your metadata, you need to configure the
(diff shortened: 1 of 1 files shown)


Resources/doc/config.rst
```diff
@@ -306,15 +306,13 @@ It is possible to connect to several mongodb servers on one connection if
 you are using a replica set by listing all of the servers within the connection
 string as a comma separated list.
 
-.. configuration-block::
```
malarzm: Why are we ditching `configuration-block`?

mhor: IMO this is an useless block, we had only yaml under this block

malarzm: I think I'd rather keep them and add other configurations instead (in another PR)

mhor: I have reverted changes for  `configuration-block`

malarzm:
Thanks @mhor!

PR #345, opened 2016-02-10
[3.2] Custom Collections

malarzm:
Exposes configuration for [custom collection classes](https://github.com/doctrine/mongodb-odm/pull/1219)


CacheWarmer/PersistentCollectionCacheWarmer.php
```diff
@@ -0,0 +1,90 @@
...
+        if (! file_exists($collCacheDir)) {
+            if (false === @mkdir($collCacheDir, 0775, true)) {
+                throw new \RuntimeException(sprintf('Unable to create the Doctrine persistent collection directory (%s)', dirname($collCacheDir)));
+            }
+        } else if (! is_writable($collCacheDir)) {
+            throw new \RuntimeException(sprintf('Doctrine persistent collection directory (%s) is not writable for the current system user.', $collCacheDir));
+        }
+
+        // if persistent collection are autogenerated we don't need to generate them in the cache warmer.
+        if (Configuration::AUTOGENERATE_NEVER !== $this->container->getParameter('doctrine_mongodb.odm.auto_generate_persistent_collection_classes')) {
```
alcaeus: Isn't the purpose of the cache warmer to prepare the cache so it won't generate cache misses during runtime? We should still generate those cached collection when the user is requesting them to be generated using the cache warmer.

malarzm: User can still generate proxies/hydrators/custom collections with specific commands, this is also how the rest of cache warmers are working so for now I'd leave it as is :)

alcaeus:
🚀 LGTM

PR #352, opened 2016-03-02
Changed toolbar design to 2.8+ style

snagels:
This pull request will make the profiler debug toolbar extension more accurate to current symfony design


malarzm:
@snagels thanks for the PR! Could we leave both classes as we still support 2.7 or does it break things?

snagels:
@malarzm This would break the design again, but there is a workaround. There is a variable you can ask for the version. I will implement this and update the pull request.

malarzm:
Great, thanks a bunch @snagels!

stof:
This is no enough to support the new profiler design fully. Icons should use SVG now (see DoctrineBundle for instance)

malarzm:
@snagels would you mind taking care of icons @stof is pointing out?

snagels:
@malarzm I will take care of it as soon as possible.

malarzm:
:+1:

PR #349, opened 2016-02-13
replaced app/console with bin/console

roman-1983:

PR #354, opened 2016-03-04
Update index.rst

cafernandez:
In symfony 3  it is bin/console not app/console


malarzm:
Thanks for the PR @cafernandez, we have #349 replacing few occurrences more that I will merge as soon as we make two version of documentation in Sf docs :)

PR #356, opened 2016-03-17
Update index.rst

marvin-SL:
Fixed a typo in documentation :  changed `app/console` to `bin/console` according to SF3


alcaeus:
Thanks! This is already being done in #349 and will be merged shortly.

PR #358, opened 2016-03-23
Load mapping files with bundle's extension

### LINKED ISSUE #357: Documentation is specifying incorrect yaml file suffix
The documentation both [here](http://symfony.com/doc/current/bundles/DoctrineMongoDBBundle/index.html#add-mapping-information) and [here](http://symfony.com/doc/current/bundles/DoctrineMongoDBBundle/config.html#mapping-configuration) says to suffix yaml mapping files with `.mongodb.yml` but this errors with `Bundle AcmeStoreBundle does not contain any mapped documents.Did you maybe forget to define a mapping configuration?`

Judging by [this code](https://github.com/doctrine/mongodb-odm/blob/master/lib/Doctrine/ODM/MongoDB/Mapping/Driver/SimplifiedYamlDriver.php#L31) it looks like the expected suffix is actually `.mongodb-odm.yml` and when I change to that the file is mapped properly.

Are those docs in a repo somewhere? Can't remember.

malarzm:
Closes #357. The bundle's drivers are still to be deprecated


malarzm:
Closing in favour of #361

PR #361, opened 2016-04-06
Correction for mapping driver file extension

caciobanu:
Fix for https://github.com/doctrine/DoctrineMongoDBBundle/issues/357

Mention: I did not remove the deprecated status of the two mapping drivers from the bundle. Should I remove the deprecated messages too ?

### DIFF
--- a/Mapping/Driver/XmlDriver.php
+++ b/Mapping/Driver/XmlDriver.php
@@ -14,8 +14,6 @@
 
 namespace Doctrine\Bundle\MongoDBBundle\Mapping\Driver;
 
-@trigger_error('The '.__NAMESPACE__.'\XmlDriver class is deprecated since version 3.1 and will be removed in 4.0. Use the Doctrine\ODM\MongoDB\Mapping\Driver\SimplifiedXmlDriver class instead.', E_USER_DEPRECATED);
-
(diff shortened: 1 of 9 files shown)


alcaeus:
> I did not remove the deprecated status of the two mapping drivers from the bundle. Should I remove the deprecated messages too ?

Since we're using them again I would no longer consider them deprecated. Also, the commit removing unit tests for those drivers (953fad5eb3ad80daa00efc9e282f39b0a3ff08d6) should also be reverted.

In the long run it would be nice if we could fix this upstream in DoctrineBridge so we don't have to add classes just to change the name of a file extension.

caciobanu:
Indeed, the fix should be moved to DoctrineBridge but I think it will be much harder to make any changes there because it will impact the ORM too.

Is it OK to add back the unit tests and remove the deprecated messages or you will do a revert ?

alcaeus:
Go ahead and add them back in while you're at it :)

caciobanu:
@alcaeus done.

caciobanu:
@alcaeus is everything good ? :)

malarzm:
Thank you @caciobanu very much for looking into this!

caciobanu:
Thank you for your time!

PR #365, opened 2016-05-13
Missing parameter when install

ihsanudin:
[Symfony\Component\DependencyInjection\Exception\ParameterNotFoundException]  
  You have requested a non-existent parameter "mongodb_server".

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -92,11 +92,18 @@ the MongoDB ODM across your application:
 
 .. code-block:: yaml
 
+    # app/config/parameters.yml
+    mongodb_server: "mongodb://localhost:27017"
+
+.. note::
+
+.. code-block:: yaml
+
     # app/config/config.yml
     doctrine_mongodb:
         connections:
             default:
-                server: mongodb://localhost:27017
+                server: "%mongodb_server%"
                 options: {}
         default_database: test_database
         document_managers:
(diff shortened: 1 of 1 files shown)


malarzm:
@ihsanudin thanks for the PR but I'm not entirely sure about the purpose (esp given PR's description)?

Resources/doc/index.rst
```diff
@@ -92,11 +92,18 @@ the MongoDB ODM across your application:
 
 .. code-block:: yaml
 
+    # app/config/parameters.yml
+    mongodb_server: "mongodb://localhost:27017"
+
+.. note::
```
stof: this empty note does not make sense

stof:
@malarzm the main benefit of this PR IMO is encouraging people to configure the mongodb server in the parameters.yml file (which is not committed) instead of hardcoding it in config.yml and hoping that the same URL is valid in development and in production

malarzm:
Encouraging people is in fact good idea :+1:

ihsanudin:
When you try to install, you will get an error because this bundle require mongodb_server as parameter like my comment

ihsanudin:
I install this bundle om Symfony 3.0.6 and got that error

Resources/doc/index.rst
malarzm: @ihsanudin would you mind removing the empty note?

ihsanudin: Done

Sorry for slow response

malarzm:
Merged manually in https://github.com/doctrine/DoctrineMongoDBBundle/commit/e9d6ce2d8ab11a8804cac1adf1cae32a08d9b82e - thanks @ihsanudin!

ihsanudin:
:+1: no problem for me

PR #371, opened 2016-06-30
Expose configuration for PersistentCollectionFactory

malarzm:
We could also expose option to set `PersistentCollectionGenerator` but I'm not sure if it's good idea


alcaeus:
👍 Looks good!

PR #378, opened 2016-09-16
Doc example update/fix

branislavmikic:
Sort example fix.


malarzm:
@branislavmikic thank you very much! :)

Issue #251, opened 2014-05-20
improve documentation on symfony.com to clarify configuration


cordoval:
related https://github.com/symfony/symfony-docs/issues/3854

currently there are two recommendations in the documentation, provide a server whole line defining configuration connection options and provide individual settings via option username and password and default_database, this last value is a bit misleading as I spent some good amount of time and it wouldn't validate correctly, I had to pass the database at the end of server: value.

I think this problem can expand to a further improvement in configuration so this is the spearhead only but we have to start somewhere.

PR #379, opened 2016-10-05
Add connection authSource option

sterrien:
We were not able to set authSource option using this bundle.

See https://github.com/doctrine/DoctrineMongoDBBundle/issues/251#issuecomment-43958755


sterrien:
What kind of test do you mean ?
Because the `options` array where the `authSource` will be (or not : `->ifNull()->thenUnset()`) will be used - as an array here : https://github.com/doctrine/DoctrineMongoDBBundle/blob/master/DependencyInjection/DoctrineMongoDBExtension.php#L293
- and especially as argument of the constructor of the connection object here : https://github.com/doctrine/mongodb/blob/master/lib/Doctrine/MongoDB/Connection.php#L79
- and finally it will be passed in the MongoClient constructor : https://github.com/doctrine/mongodb/blob/master/lib/Doctrine/MongoDB/Connection.php#L282

By comparing the symfony documentation (http://symfony.com/doc/master/bundles/DoctrineMongoDBBundle/config.html#full-default-configuration) and the php.net page (http://php.net/manual/en/mongoclient.construct.php), I found 4 missing options :
- authSource
- fsync
- gssapiServiceName
- secondaryAcceptableLatencyMS

Do you want I add those 4 options ?

(I also read that we should use MongoDB\Driver\Manager class instead of MongoClient, but that's another story)

alcaeus:
Sorry for the lack of a better explanation before. There is a test for the configuration class which tests the full option set available: https://github.com/doctrine/DoctrineMongoDBBundle/blob/master/Tests/DependencyInjection/ConfigurationTest.php#L55. This test should include the new option to make sure it's not accidentally removed without a test failing.

As for the new options, if you could add them while you're at it, that'd be helpful 👍 

> I also read that we should use MongoDB\Driver\Manager class instead of MongoClient, but that's another story

That will be fixed in ODM 2.0 as the old driver (which provides MongoClient) and the new driver are not API compatible.

alcaeus:
Superseded by #396.

PR #381, opened 2016-11-10
Code cleanup for PHP 5.6

alcaeus:
This PR cleans up code since we're using PHP 5.6 and newer:
* PHPUnit is now included in the dev dependencies, this should make development easier
* Instead of having classes as string literals, we use the `::class` constant
* Arrays have been converted to short syntax
* With PHP 7.1 coming up, it's time to add it to the build matrix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,3 +27,10 @@ before_install:
 
 install:
   - composer update $COMPOSER_FLAGS
+
+script:
+    - ./vendor/bin/phpunit --coverage-clover=coverage.clover
+
+after_script:
+  - wget https://scrutinizer-ci.com/ocular.phar
+  - php ocular.phar code-coverage:upload --format=php-clover coverage.clover
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,20 @@
 language: php
+sudo: false
 
 php:
   - 5.6
+  - 7.0
+  - 7.1
 
 env:
-  - MONGO_DRIVER=mongo
+  global:
+    - DRIVER_VERSION="stable"
+    - ADAPTER_VERSION="^1.0.0"
 
 matrix:
   include:
     - php: 5.6
-      env: MONGO_DRIVER=mongo COMPOSER_FLAGS='--prefer-lowest --prefer-stable'
-    - php: 7.0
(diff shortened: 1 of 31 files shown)


stof:
this branch already has conflicts

alcaeus:
@stof Yes, noticed I had stale branches on my machine here. I've rebased it against master, so we should be good now.

.travis.yml
```diff
@@ -1,29 +1,37 @@
...
 
 matrix:
   include:
     - php: 5.6
-      env: MONGO_DRIVER=mongo COMPOSER_FLAGS='--prefer-lowest --prefer-stable'
-    - php: 7.0
-      env: ADAPTER_VERSION="^1.0.0" MONGO_DRIVER=mongodb
-
-sudo: false
+      env: DRIVER_VERSION="1.5.8" PREFER_LOWEST="--prefer-lowest"
```
stof: naming it ``COMPOSER_FLAGS`` is better than ``PREFER_LOWEST``, as it allows to pass any flags to the composer command

.travis.yml
```diff
@@ -1,29 +1,37 @@
...
-  - yes '' | pecl -q install -f $MONGO_DRIVER
-  - if [ "x${ADAPTER_VERSION}" != "x" ]; then composer require "alcaeus/mongo-php-adapter=${ADAPTER_VERSION}" --ignore-platform-reqs; fi
-
 install:
-  - composer update $COMPOSER_FLAGS
+  - composer self-update
+  - if [[ ${TRAVIS_PHP_VERSION:0:2} == "5." ]]; then yes '' | pecl -q install -f mongo-${DRIVER_VERSION}; fi
+  - if [[ ${TRAVIS_PHP_VERSION:0:2} == "7." ]]; then pecl install -f mongodb-${DRIVER_VERSION}; fi
+  - if [[ ${TRAVIS_PHP_VERSION:0:2} == "7." ]]; then composer require "alcaeus/mongo-php-adapter=${ADAPTER_VERSION}" --ignore-platform-reqs; fi
+  - composer update ${PREFER_LOWEST}
```
stof: you should add ``--prefer-dist``, which is better for caching (now that you use the container-based system which has access to it)

alcaeus:
@stof Thanks for the feedback, I've opened #382 with the changes.

PR #383, opened 2016-11-18
A few minor tweaks in docs

bocharsky-bw:


malarzm:
Thanks @bocharsky-bw!

PR #384, opened 2016-11-18
Reorder supported field types by name ASC

bocharsky-bw:


malarzm:
And thanks again @bocharsky-bw :)

PR #387, opened 2016-11-30
/DataCollector/PrettyDataCollector.php check $log

muttley73:
Hi,
in line 101

`
if ($log['query'] || $log['limit'] $log['skip']) 
`

create error on php 5.6.28 version ( index "query" do not exist)

i have modified with:

`
if (isset($log['query']) || isset($log['limit']) || isset($log['skip'])) {
`


malarzm:
@muttley73 thanks for the PR! Would you mind creating it with your changes only though?

PR #389, opened 2016-12-01
Fix tests when running with Symfony 3.2

alcaeus:
It seems with Symfony 3.2 the `kernel.name` parameter is no longer automagically set in the container. To solve this, we declare it in all containers we're building. This PR also fix a deprecation warning about a duplicate key in the yaml configuration used in the tests.

Note: we might want to backport this PR to the 3.0 branch - might do that later.


malarzm:
Oh, wanted to investigate this during hackday tomorrow, thanks for doing so sooner :D

PR #390, opened 2016-12-05
use isset() in PrettyDataCollector class (rebase for test)

muttley73:


alcaeus:
FWIW, you don't need to open a new pull request every time you have to rebase. You can use `git rebase` and `git push --force` to overwrite commits in the old branch.

muttley73:
thanks @alcaeus

alcaeus:
@muttley73 While checking out tests for your fix I've noticed that the count queries we generated weren't quite valid anyways. I've adressed this in #391, along with the checks you've added in this PR. Thanks for getting us started on a fix!

PR #393, opened 2016-12-27
Allow null value for replicaSet

caciobanu:
This is useful to use the same configuration to connect to MongoDB with no replica set configured or with replica set configured and easily switchable by a parameter like so:

```code
doctrine_mongodb:
    ...
    connections:
        default:
            server: %mongo_connection%
            options:
                connect: true
                replicaSet: %mongo_connection_replica_set%
```

In DEV where I have no replica set I can set the parameter **mongo_connection_replica_set** to **null** and in PROD where I have replica set I can set it to the replica set name.


malarzm:
Sweet, I remember I switched to using DSN because of this few years ago :+1:

Steveb-p:
@malarzm Just to clarify, this is merged into 3.2 tag, right?

alcaeus:
No - according to the commit the first tag the commit appears in is 3.3.0. Not sure why I merged into 3.2 branch without having a 3.2 release, but the milestone information is correct on this one.

PR #405, opened 2017-03-15
Update config.srt

KoenraadM:
Authentication documentation seems to be out of date

### DIFF
--- a/Resources/doc/config.rst
+++ b/Resources/doc/config.rst
@@ -408,7 +408,7 @@ Otherwise you will get a *auth failed* exception.
                     options:
                         username: someuser
                         password: somepass
-                        db: db_you_have_access_to
+                        authSource: db_you_have_access_to
 
     .. code-block:: xml
 
--- a/Resources/doc/config.rst
+++ b/Resources/doc/config.rst
@@ -405,9 +405,10 @@ Otherwise you will get a *auth failed* exception.
             connections:
                 default:
                     server: "mongodb://localhost:27017"
-                    username: someuser
-                    password: somepass
-                    authSource: db_you_have_access_to
+                    options:
+                        username: someuser
+                        password: somepass
+                        db: db_you_have_access_to
 
     .. code-block:: xml
 


Resources/doc/config.rst
```diff
@@ -405,9 +405,10 @@ Otherwise you will get a *auth failed* exception.
             connections:
                 default:
                     server: "mongodb://localhost:27017"
-                    username: someuser
-                    password: somepass
-                    authSource: db_you_have_access_to
+                    options:
+                        username: someuser
+                        password: somepass
+                        db: db_you_have_access_to
```
alcaeus: This needs to be `authSource`, not `db`. There's a difference between them as indicated by the paragraph above this configuration block.

KoenraadM: Hmm, the `authSource` doesn't seem to work for me ... unless I'm doing something wrong.
`Unrecognized option "authSource" under "doctrine_mongodb.connections.app.options"  `

alcaeus: We're having an old problem: the documentation is generated from the master branch, which isn't released yet. I'd assume you're using a stable (3.2) release which doesn't yet have the `authSource` option. We could add a note about the `authSource` option being introduced in an upcoming release.

KoenraadM: Even in the fixtures of 3.2.1 (at least, I haven't checked any other versions) I can't seem to find `authSource`

```
connections:
        conn1:
            server:               mongodb://localhost
            options:
                connect:          true
                connectTimeoutMS: 500
                db:               database_val
                journal:          true
                password:         password_val
                readPreference:   secondaryPreferred
                readPreferenceTags:
                  - { dc: east, use: reporting }
                  - { dc: west }
                  - {  }
                replicaSet:       foo
                slaveOkay:        true
                socketTimeoutMS:  1000
                ssl:              true
                authMechanism:    X509
                username:         username_val
                w:                majority
                wTimeoutMS:       1000
        conn2:
            server: mongodb://otherhost
```

alcaeus: Yes, that's what I said: the current stable release line (3.2) does not yet have the option, it will be introduced in 3.3 once we release that.

KoenraadM: Ah, so if I just change this to authSource, and PR this into master, everything will be fine when master is released? Changing `db` in `authSource` then :D

KoenraadM: But, does that then imply that the current options (user, pass, authSource) is pulled to the parent level too, or am I misunderstanding this?

alcaeus: Yes, the documentation will be correct for master. The documentation was incorrect, even before this was changed in master. 3.2 also has `username`, `password` and other options under an `options` key.

PR #406, opened 2017-04-10
Format aggregation queries in data collector

### LINKED ISSUE #392: Show aggregation queries in data collectors
This was initially reported in https://github.com/doctrine/mongodb-odm/issues/1534. The data collectors don't handle aggregation queries run either through `$collection->aggregate()` or the aggregation builder.

alcaeus:
Fixes #392.

### DIFF
--- a/DataCollector/PrettyDataCollector.php
+++ b/DataCollector/PrettyDataCollector.php
@@ -84,7 +84,15 @@ class PrettyDataCollector extends StandardDataCollector
                 }
 
                 // format the method call
-                if (isset($log['authenticate'])) {
+                if (isset($log['aggregate'])) {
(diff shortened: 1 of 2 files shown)


DataCollector/PrettyDataCollector.php
```diff
@@ -84,7 +84,9 @@ public function collect(Request $request, Response $response, \Exception $except
                 }
 
                 // format the method call
-                if (isset($log['authenticate'])) {
+                if (isset($log['aggregate'])) {
+                    $query .= '.aggregate('.$this->bsonEncode($log['pipeline']).', '.$this->bsonEncode($log['options']).')';
```
malarzm: I think `options` should be included only when not empty to avoid having `[  ]` in logged query

PR #409, opened 2017-04-28
Prepare 3.3 release

malarzm:


composer.json
```diff
@@ -39,7 +39,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "3.1-dev"
+            "dev-master": "3.3-dev"
```
alcaeus: IMO it would make sense to branch off 3.3 and change dev-master to `3.4-dev`.

stof: not sure about it. Do you really plan to maintain several branches ? For a bundle, it is probably fine to have a single maintained branch (except for multiple major version)

alcaeus: Sorry for the late reply. In the past, we've created a branch to support stable releases while making master the development branch for the next minor or major version, depending on the changes.

PR #410, opened 2017-05-03
Change Twig notation for the templates

Lctrs:
Playing around with a Symfony project created with Symfony Flex, I noticed that I got the error below everytime I try to access a profile in the profiler.

`The profiler template "DoctrineMongoDBBundle:Collector:mongodb.html.twig" for data collector "mongodb" does not exist.`

Changing the Twig notation fixed the problem.


Lctrs:
This notation is already used somewhere else in the Twig template. So there shouldn't be any problem for the other supported versions of Symfony.

malarzm:
Thanks @Lctrs!

PR #413, opened 2017-05-18
Allow Symfony 4.0

malarzm:


malarzm:
Re 2.7 - Given it's not causing any maintenance overhead so far I'd drop it when we do bundle for ODM 2.0 :)

PR #415, opened 2017-06-02
Doc - Fix compatability with Symfony 3.x

michaelperrin:
The documented example didn't work with Symfony 3.x (leading to a new user who didn't understand why these examples didn't work on the Symfony support slack channel).

I fixed the examples that now work for sure with Symfony 2.8+ and Symfony 3. The removal of the `getName()` method may introduce an incompatibility with Symfony 2.7 (not sure about it), but I don't think this is a problem. Better make it work on Symfony 2.8+ rather than on 2.7 and 2.8 only.


alcaeus:
Test failures are unrelated, deprecation in Symfony 3.3 causes tests to appear failed. Thanks @michaelperrin!

PR #412, opened 2017-05-15
Note explaining confunding drivers error for php7

gustawdaniel:
Sources:

> https://docs.mongodb.com/ecosystem/drivers/php/

> http://stackoverflow.com/questions/36984973/php70-mongo-install-doctrine-mongodb-odm-fails

I lost a lot of time looking for hint like this. So I hope that this note will help other programmers.

PR #417, opened 2017-06-02
Add a note on PHP 7

malarzm:
Supersedes #412

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -57,6 +57,12 @@ command from the directory where your ``composer.json`` file is located:
 Now, Composer will automatically download all required files, and install them
 for you.
 
+.. note::
+
+   Currently ODM works with legacy MongoDB driver (i.e. `ext-mongo`) which is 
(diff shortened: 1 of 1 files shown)


Resources/doc/index.rst
```diff
@@ -57,6 +57,12 @@ command from the directory where your ``composer.json`` file is located:
 Now, Composer will automatically download all required files, and install them
 for you.
 
+.. note::
+
+   Currently ODM works with legacy MongoDB driver (i.e. `ext-mongo`) which is 
+   not compatible with PHP 7. If you have PHP 7 and/or are using new driver 
+   (i.e. 1ext-mongodb1) please refer to `"Using PHP 7" section`_ in ODM's documentation.
```
alcaeus: I think you meant to type a single \`, not a `1`, right?

malarzm: That's right ;)

malarzm:
@alcaeus good catch, fixed now!

PR #418, opened 2017-06-02
Fix form creation for latest versions of Symfony

michaelperrin:
Sorry, there was one last thing to change in the documentation for compatibility with Symfony 2.8+, so here is an other PR (I couldn't squash commits as #415 is already merged).


malarzm:
Thanks for the followup @michaelperrin!

PR #421, opened 2017-06-20
WIP: Add explain() and command stats to debug panel

xosofox:
**This is work in progress, requesting inputs/comments**

Goal: Give execution times, execution plan information, index scan info etc in debug panel

Todo/to discuss
- [ ] Unit tests? How? What does make sense?
- [ ] hackish JS show/hide. Works, but...
- [ ] add all(?)/more post-hooks for commands (aggregate, findAndUpdate, findAndRemove, ...)

Here some screenshots to see what it currently looks like

toolbar time
query list
query list explain details


malarzm:
Out of curiosity: what is the performance impact of running `explain` for every query?

Xosofox:
@malarzm This actually is a good point!! Seems to be significant:

On an arbitrary 21 requests, the total time is a lot higher! So it looks like an "opt-in" would make sense. I defaulted the explain value to ```%kernel.debug%``` right now, but false might be a better option?

no_explain

with_explain

Can of course be changed here:

```
document_managers:
        default:
            auto_mapping: true
            profiler:
              explain: false
```

malarzm:
Not sure if possible/doable (or even secure, as @jmikola hinted) but is it possible to run explain on demand? Having an endpoint that explains arbitrary queries is no good, but executing explain on 3rd query of profile `xyz` seems safe at a glance. Practically "Explain" would fire an ajax request to an endpoint residing in profiler (so not reachable with profiler disabled) and show returned data. What do you think?

Xosofox:
Running explain "afterwards" might give you different results than when the query in question was actually executed. So, to prevent that, you would still have to run it in the background every time, store the results and only show them on demand - and there is neither performance nor usability benefit for that.

malarzm:
> Running explain "afterwards" might give you different results than when the query in question was actually executed.

I was hoping that few minutes between saving the profile and inspecting it will not matter much but I see your point. That leaves us with your proposition to have `false` by default and explicitly opting in for explaining all queries. 

Offhand does ORM bundle has something alike for analyzing queries?

Xosofox:
>Offhand does ORM bundle has something alike for analyzing queries?

Yes, actually very good, that's where I stole parts of the styles from

malarzm:
Do they do the same for analyzing? I mean collect all queries, analyze and store its results?

DataCollector/StandardDataCollector.php
```diff
@@ -54,8 +55,66 @@ public function getQueries()
...
+        $c = count($this->queries);
+        if (0 === $c) {
+            return;
+        }
+
+        $i = $c - 1;
+        $cursor = $args->getData();
+        try {
+            $explain = $cursor->explain();
+            dump($explain);
```
WalentinG: Dump?

Issue #422, opened 2017-08-14
Why doctrine/mongodb-odm is required in installation instructions


murilolobato:
Reading the installation instructions on [symfony.com](http://symfony.com/doc/current/bundles/DoctrineMongoDBBundle/index.html#installation) and in the own [repo docs](https://github.com/doctrine/DoctrineMongoDBBundle/blob/master/Resources/doc/index.rst#installation ) (which is referenced by the symfony webpage) it says to add the ```doctrine/mongodb-odm``` to the composer.json file. 

Why is this necessary? Isn't ```doctrine/mongodb-odm-bundle``` enough, since it already depends on ```doctrine/mongodb-odm``` ?

alcaeus:
I'm guessing it's left over from when the `@dev` suffix (see https://github.com/doctrine/DoctrineMongoDBBundle/commit/16f518005d53c4f8e1cb7c3de9f7c55309a5c50b) was dropped. Previously, that was necessary because there wasn't a stable release of ODM, so composer requires you to explicitly specify the dependency in the root composer.json. If you want you can create a pull request to simplify this.

alcaeus:
#423 has been merged, closing here.

PR #423, opened 2017-08-14
Updated installation instructions

murilolobato:
Update installation instructions to contain only a dependency to ```doctrine/doctrine-odm-bundle```.

Related to #422


alcaeus:
Thanks!

PR #425, opened 2017-09-07
Add Shard command from DoctrineODM

Steveb-p:
This patch adds `odm:schema:shard` command from DoctrineODM library.

See: http://docs.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/sharding.html


malarzm:
Thanks a bunch @Steveb-p!

Issue #426, opened 2017-09-08
Bad namespace of ShardCmd


hefekranz:
Hey guys, there seems to be a mistake in the namespace of this new [Shard command](https://github.com/doctrine/DoctrineMongoDBBundle/blob/master/Command/ShardDoctrineODMCommand.php)

`namespace DoctrineMongoDBBundle\Command;`

should be

`namespace Doctrine\Bundle\MongoDBBundle\Command;`

Reproducable with:

```
FROM debian:jessie

RUN apt-get update && apt-get install -y curl git mongodb-server php5-cli php5-mongo
RUN service mongodb start
RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer

RUN mkdir -p /usr/local/bin \
    && curl -LsS https://symfony.com/installer -o /usr/local/bin/symfony \
    && chmod a+x /usr/local/bin/symfony \
    && symfony new vanilla

WORKDIR vanilla

RUN composer require doctrine/mongodb-odm-bundle dev-master && composer update

RUN echo 'doctrine_mongodb:\n\
    connections:\n\
        default:\n\
            options: {}\n\
    default_database: test_database\n\
    document_managers:\n\
        default:\n\
            auto_mapping: true\n'>> app/config/config.yml

RUN sed -i 's/AppBundle\\AppBundle(),/Doctrine\\Bundle\\MongoDBBundle\\DoctrineMongoDBBundle()/g' app/AppKernel.php

RUN php bin/console
```

hefekranz:
`  [RuntimeException]                                                           
  The autoloader expected class "Doctrine\Bundle\MongoDBBundle\Command\ShardD  
  octrineODMCommand" to be defined in file "/vanilla/vendor/composer/../doctr  
  ine/mongodb-odm-bundle/Command/ShardDoctrineODMCommand.php". The file was f  
  ound but the class was not in it, the class name or namespace probably has   
  a typo.`

malarzm:
Just merged #427 so I'm closing the report - thank you very much for noticing and I'm sorry for overlooking this originally :)

PR #427, opened 2017-09-08
fix for bad namespace in ShardDoctrineODMCommand

hefekranz:
#426


malarzm:
@hefekranz you are absolutely right, thank you very much for the fix!

hefekranz:
my pleasure, cheers

PR #433, opened 2017-10-30
Fix Symfony 4 BC breaks

MatTheCat:
Based on deprecation warnings got from Symfony 3.4:

- The "**Symfony\Bridge\Doctrine\ManagerRegistry::setContainer()**" method is deprecated since version 3.4 and will be removed in 4.0. Inject a PSR-11 container using the constructor instead.
- The "**doctrine_mongodb**" service is private, getting it from the container is deprecated since Symfony 3.2 and will fail in 4.0. You should either make the service public, or stop using the container directly and use dependency injection instead.
- The "**doctrine_mongodb.odm.default_document_manager**" service is private, getting it from the container is deprecated since Symfony 3.2 and will fail in 4.0. You should either make the service public, or stop using the container directly and use dependency injection instead.
- Implementing "**Symfony\Component\HttpKernel\DataCollector\DataCollectorInterface**" without the "**reset()**" method is deprecated since version 3.4 and will be unsupported in 4.0 for class "**Doctrine\Bundle\MongoDBBundle\DataCollector\PrettyDataCollector**".

### DIFF
--- a/DataCollector/StandardDataCollector.php
+++ b/DataCollector/StandardDataCollector.php
@@ -44,6 +44,14 @@ class StandardDataCollector extends DataCollector implements LoggerInterface
         $this->data['queries'] = array_map('json_encode', $this->queries);
     }
 
+    public function reset()
+    {
+        $this->data = [
+            'nb_queries' => 0,
+            'queries' => [],
+        ];
+    }
+
     public function getQueryCount()
     {
         return $this->data['nb_queries'];
--- a/DataCollector/StandardDataCollector.php
+++ b/DataCollector/StandardDataCollector.php
@@ -46,6 +46,7 @@ class StandardDataCollector extends DataCollector implements LoggerInterface
 
     public function reset()
     {
+        $this->queries = [];
         $this->data = [
             'nb_queries' => 0,
             'queries' => [],
--- a/ManagerRegistry.php
+++ b/ManagerRegistry.php
@@ -15,16 +15,24 @@
 namespace Doctrine\Bundle\MongoDBBundle;
 
 use Doctrine\ODM\MongoDB\MongoDBException;
-use Psr\Container\ContainerInterface;
 use Symfony\Bridge\Doctrine\ManagerRegistry as BaseManagerRegistry;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 
 class ManagerRegistry extends BaseManagerRegistry
 {
     public function __construct(ContainerInterface $container, $name, array $connections, array $managers, $defaultConnection, $defaultManager, $proxyInterfaceName)
     {
-        parent::__construct($name, $connections, $managers, $defaultConnection, $defaultManager, $proxyInterfaceName);
+        $parentTraits = class_uses(parent::class);
+        if (isset($parentTraits[ContainerAwareTrait::class])) {
+            // this case should be removed when Symfony 3.4 becomes the lowest supported version
+            // and then also, the constructor should type-hint Psr\Container\ContainerInterface
+            $this->setContainer($container);
+        } else {
+            $this->container = $container;
+        }
 
-        $this->container = $container;
+        parent::__construct($name, $connections, $managers, $defaultConnection, $defaultManager, $proxyInterfaceName);
     }
 
     /**
--- a/ManagerRegistry.php
+++ b/ManagerRegistry.php
@@ -15,10 +15,18 @@
 namespace Doctrine\Bundle\MongoDBBundle;
 
 use Doctrine\ODM\MongoDB\MongoDBException;
+use Psr\Container\ContainerInterface;
 use Symfony\Bridge\Doctrine\ManagerRegistry as BaseManagerRegistry;
 
 class ManagerRegistry extends BaseManagerRegistry
 {
+    public function __construct(ContainerInterface $container, $name, array $connections, array $managers, $defaultConnection, $defaultManager, $proxyInterfaceName)
+    {
+        parent::__construct($name, $connections, $managers, $defaultConnection, $defaultManager, $proxyInterfaceName);
+
+        $this->container = $container;
+    }
+
     /**
      * Resolves a registered namespace alias to the full namespace.
      *
--- a/ManagerRegistry.php
+++ b/ManagerRegistry.php
@@ -21,7 +21,7 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
 
 class ManagerRegistry extends BaseManagerRegistry
 {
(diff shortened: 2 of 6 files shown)


ManagerRegistry.php
```diff
@@ -15,10 +15,18 @@
...
+use Psr\Container\ContainerInterface;
 use Symfony\Bridge\Doctrine\ManagerRegistry as BaseManagerRegistry;
 
 class ManagerRegistry extends BaseManagerRegistry
 {
+    public function __construct(ContainerInterface $container, $name, array $connections, array $managers, $defaultConnection, $defaultManager, $proxyInterfaceName)
+    {
+        parent::__construct($name, $connections, $managers, $defaultConnection, $defaultManager, $proxyInterfaceName);
+
+        $this->container = $container;
```
stof: this won't work with older versions of Symfony. Look at DoctrineBundle for the way to support both

DataCollector/StandardDataCollector.php
```diff
@@ -44,6 +44,14 @@ public function collect(Request $request, Response $response, \Exception $except
         $this->data['queries'] = array_map('json_encode', $this->queries);
     }
 
+    public function reset()
+    {
+        $this->data = [
```
stof: this need to reset `$this->queries` too

MatTheCat: Done

ManagerRegistry.php
MatTheCat: Done

ManagerRegistry.php
```diff
@@ -16,9 +16,25 @@
 
 use Doctrine\ODM\MongoDB\MongoDBException;
 use Symfony\Bridge\Doctrine\ManagerRegistry as BaseManagerRegistry;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 
 class ManagerRegistry extends BaseManagerRegistry
 {
+    public function __construct(ContainerInterface $container, $name, array $connections, array $managers, $defaultConnection, $defaultManager, $proxyInterfaceName)
```
alcaeus: Isn’t this signature change a BC break, requiring a major release?

MatTheCat: Hm yes I guess it would break if someone relies on arguments index to override the definition. Could we rely on setter injection instead?

alcaeus: You can add the `ContainerInterface` at the end and change its default value to `null`. That way it's not a BC break.

MatTheCat: Oh right

pierre-tranchard:
I don't know about you, but if I apply your modification to get the bundle works, it does but I don't have access to any command like if they were not registered to the application. 
Thanks for proposing your PR to make the bundle compatible by the way

MatTheCat:
@pierre-tranchard cannot reproduce neither on 3.4 nor on 4.0

michaelperrin:
I didn't realize you already opened this PR @MatTheCat when I started work on the same thing.
I closed my PR (#436) as a result.

Maybe you could add two things I introduced in mine to yours:

* Mark `doctrine_mongodb.odm.document_manager` service as public (see https://github.com/doctrine/DoctrineMongoDBBundle/pull/436/files#diff-21bb4bdcf714a43b98948d9c96f74a1fR274 ).
* Add a test to check that services are public (see https://github.com/doctrine/DoctrineMongoDBBundle/pull/436/files#diff-14dd137c42215df91c67dc25d9dba7ca).
* Add a test for the `reset` method: https://github.com/doctrine/DoctrineMongoDBBundle/pull/436/files#diff-ccb5be2343457e6434328c4ed8582c83 (it needs to be upgraded as your `reset` method implementation is more complete than mine)

pierre-tranchard:
This maybe useful too for Symfony 4
https://github.com/doctrine/DoctrineMongoDBBundle/pull/437

malarzm:
Also if you could rebase atop current `master` branch that would make the deprecation-related failures go away :)

MatTheCat:
@pierre-tranchard thanks! Dunno why I didn't get the issue.

@michaelperrin isn't the alias of a public service also public? Also are these tests really required?

@malarzm will do tomorrow!

pierre-tranchard:
cool, if we can merge it tomorrow after @stof approval it would be awesome 👍

alcaeus:
> isn't the alias of a public service also public? Also are these tests really required?

A service alias inherits the `public` setting from the `defaults` of the config file, not whether the original service is public. Thus, the alias has to be marked as public for this to work in Symfony 4.

MatTheCat:
Isn't it what we want then? Deprecations come from this bundle because it accesses services through the container but aliases are meant to be used by the end user right? So it would be his choice to make it public or not.

alcaeus:
If you want the alias to be public, mark it as such. If you want it to be private, mark it as such. Everything else will cause inconsistent behavior between Symfony versions.

michaelperrin:
@MatTheCat I think that a test for the the `reset` method can be useful. Concerning the ones that test that services are public, I took inspiration from this PR equivalent for the ORM. I think that these tests ensure the bundle works well across Symfony's versions and can therefore be useful too.

MatTheCat:
@alcaeus the only inconsistency here is because this bundle access services which can be private though the container so I made these services public. Now I don't understand why we should chose between a public or private alias as the bundle don't access it.

@michaelperrin okay will add your tests

alcaeus:
The inconsistency comes from services being public by default in Symfony 2 and 3 while being private in Symfony 4. That's why I'm suggesting to explicitly marking it as public or private.

MatTheCat:
This is only an inconsistency for services this bundle access through the container because then they need to be public. If this bundle don't access `doctrine_mongodb.odm.document_manager` through the container we don't care if it's public or not. If the developper wants it public he will define it as public and it will have nothing to do with this bundle.

But in fact I see `TailCursorDoctrineODMCommand` get `doctrine_mongodb.odm.document_manager` through the container which seems weird to me but if it's correct we indeed have to set it public.

alcaeus:
> If this bundle don't access `doctrine_mongodb.odm.document_manager` through the container we don't care if it's public or not.

I disagree. The service was public before, allowing anyone to fetch it from the container as they see fit. Making it private is a BC break which should only be made in a major release, not a patch release.

MatTheCat:
I see your point, anyways I needed to make it explicitly public in order to avoid breaking `TailCursorDoctrineODMCommand` so it seems we're good?

pierre-tranchard:
@MatTheCat @stof @alcaeus Can we merge it by tonight if everything's well for you? I would need it for a project :)

MatTheCat:
I'm not even contributor so I can't merge it. @stof seems you're locking two of my PRs could you take a look please?

alcaeus:
I'll have some time to look at this later today.

alcaeus:
Thanks @MatTheCat!

PR #435, opened 2017-11-10
Use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase

gabriel-caruso:
I use the `PHPUnit\Framework\TestCase` notation instead of `PHPUnit_Framework_TestCase` while extending our TestCases. This will help us migrate to PHPUnit 6, that [no longer support snake case class names](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

I just need to bump PHPUnit version to [`^5.7`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-5.7.md#570---2016-12-02), that support this namespace.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "symfony/form": "^2.7 || ^3.0 || ^4.0",
         "symfony/yaml": "^2.7 || ^3.0 || ^4.0",
         "symfony/phpunit-bridge": "^2.7 || ^3.0 || ^4.0",
-        "phpunit/phpunit": "^5.6"
+        "phpunit/phpunit": "^5.7"
     },
     "suggest": {
         "doctrine/data-fixtures": "Load data fixtures"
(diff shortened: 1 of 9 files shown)


composer.json
```diff
@@ -27,7 +27,7 @@
         "symfony/form": "^2.7 || ^3.0 || ^4.0",
         "symfony/yaml": "^2.7 || ^3.0 || ^4.0",
         "symfony/phpunit-bridge": "^2.7 || ^3.0 || ^4.0",
-        "phpunit/phpunit": "^5.6"
+        "phpunit/phpunit": "^5.7"
```
alcaeus: As in many other repos, this should be `^5.7|^6.4`.

gabriel-caruso: Gonna work on that!

gabriel-caruso: @alcaeus Please merge this namespaces changes now. I'll work on PHPUnit 6 support latter, a lot of stuff broken.

alcaeus:
Thanks!

PR #441, opened 2017-11-27
Add aliases for autowiring

michaelperrin:
This allows to implement services with autowiring like this, which is very handy.

```php
<?php

use Doctrine\Common\Persistence\ObjectManager;

class AcmeController
{
    public function indexAction(Request $request, ObjectManager $om)
    {
        $hikes = $om->getRepository(Hike::class)->find();
        // ...
    }
}
```

### DIFF
--- a/Resources/config/mongodb.xml
+++ b/Resources/config/mongodb.xml
@@ -56,6 +56,8 @@
     <services>
         <!-- defaults -->
         <service id="doctrine_mongodb.odm.cache" alias="doctrine_mongodb.odm.cache.array" />
+        <service id="%doctrine_mongodb.odm.document_manager.class%" alias="doctrine_mongodb.odm.document_manager" public="false" />
+        <service id="%doctrine_mongodb.odm.class%" alias="doctrine_mongodb" public="false" />
 
         <service id="Doctrine\Bundle\MongoDBBundle\Command\ClearMetadataCacheDoctrineODMCommand" class="Doctrine\Bundle\MongoDBBundle\Command\ClearMetadataCacheDoctrineODMCommand">
             <tag name="console.command"/>
@@ -92,8 +94,8 @@
         </service>
         <service id="Doctrine\Bundle\MongoDBBundle\Command\UpdateSchemaDoctrineODMCommand" class="Doctrine\Bundle\MongoDBBundle\Command\UpdateSchemaDoctrineODMCommand">
             <tag name="console.command"/>
-        </service>        
-        
+        </service>
+
         <!-- events -->
         <service id="doctrine_mongodb.odm.connection.event_manager" class="%doctrine_mongodb.odm.event_manager.class%" public="false" abstract="true">
             <argument type="service" id="service_container" />


malarzm:
Does ORM have same aliases? If yes, then we'll have a conflict

Resources/config/mongodb.xml
```diff
@@ -56,6 +56,9 @@
     <services>
         <!-- defaults -->
         <service id="doctrine_mongodb.odm.cache" alias="doctrine_mongodb.odm.cache.array" />
+        <service id="Doctrine\Common\Persistence\ObjectManager" alias="doctrine_mongodb.odm.document_manager" public="false" />
```
alcaeus: No good - other mappers may also be implementing this interface. Use the `DocumentManager` interface.

Resources/config/mongodb.xml
```diff
@@ -56,6 +56,9 @@
...
+        <service id="%doctrine_mongodb.odm.document_manager.class%" alias="doctrine_mongodb.odm.document_manager" public="false" />
+        <service id="Doctrine\Common\Persistence\ManagerRegistry" alias="doctrine_mongodb" public="false" />
```
alcaeus: This will also conflict with ORM - you can use `Doctrine\Bundle\MongoDBBundle\ManagerRegistry`.

michaelperrin:
@malarzm @alcaeus It does conflict indeed (see https://github.com/doctrine/DoctrineBundle/blob/master/Resources/config/orm.xml#L78), but it allows to use interfaces instead of implementations for autowiring when the ORM and ODM are not both used in a project.

But I can indeed make the necessary changes to use the direct implementations instead.

malarzm:
Does anybody know what Symfony will do in case of a conflict? Is it an exception or bundle that is registered later "wins"?

alcaeus:
> Does anybody know what Symfony will do in case of a conflict?

IIRC, it throws an exception.

michaelperrin:
Thanks @alcaeus for bringing this topic to the DoctrineBundle as well as a separate issue in #740 :)
I don't have much opinion on this, just like @weaverryan , so I prefer to wait for the community decision.

In any case, we will need a way to autowire the ODM services. Maybe that *for now* I can change my PR to make use of the ODM with its specific interfaces.

alcaeus:
@michaelperrin using ODM specific interfaces is a start - I won't merge a pull request that adds autowiring aliases for the common interfaces, that just opens a whole new can of worms.

Since you're adding autowiring to the bundle, would you be willing to port repository services to this bundle from its ORM counterpart (introduced in https://github.com/doctrine/DoctrineBundle/pull/727)?

michaelperrin:
@alcaeus I agree with you, I will do that as a start.

I'll look promptly into the repository services. Should I make it as a separate PR?

alcaeus:
> I'll look promptly into the repository services. Should I make it as a separate PR?

Thanks! I believe it's better to split them up: that way we can merge these aliases and add service repositories later. Please let me know if you need help with anything!

michaelperrin:
@alcaeus Sorry for the delay, I made the necessary changes as discussed. I will let you know for the other PR (probably tomorrow) and thanks for your help!

alcaeus:
Thanks @michaelperrin!

michaelperrin:
Thanks for the merge @alcaeus :)
For the doctrine/DoctrineBundle#727 counterpart, it will take a bit more time, but don't worry, I will get it done (probably next week).

kejwmen:
@alcaeus when do you expect to do next patch release? Would be great to have this tagged.

alcaeus:
@kejwmen this will not be in a patch release but rather a minor release along with service repositories (unless @michaelperrin asks for more time, in which case I'll reconsider).

kejwmen:
:ok_hand:, thanks for quick reply.

michaelperrin:
@alcaeus @kejwmen I think I will have some time by the end of the week for implementing service repositories. I will keep you posted in any case :)

pensiero:
@michaelperrin @alcaeus maybe I'm missing something but, I would like to autowire both the EntityManager and DocumentManager in the same Controller. Is it possible?

Because with the ORM is perfectly fine (I just need to specify `EntityManagerInterface` as parameter), but with ODM if I use `ObjectManager` it gives me back the `EntityManager`, instead of the `DocumentManager`.

Thanks

pensiero:
PS: side note, I'm using Symfony 4, so I can only use interfaces for auto-wiring

Majkl578:
> I'm using Symfony 4, so I can only use interfaces for auto-wiring

This is an interesting point, how did you come up with this assumption? As far as I know Symfony 4 doesn't impose any interface-only autowiring, it works perfectly fine with classes as well. It only doesn't automatically add autowiring aliases for implemented interfaces anymore: https://symfony.com/doc/current/service_container/autowiring.html#working-with-interfaces

weaverryan:
That’s mostly correct :). Autowiring works for classes or interfaces. But typically, the aliases that enable autowiring are only added for interfaces. That may make it feel like only interfaces work in Symfony 4, but that’s not true :)

I said you’re *mostly* correct only because if you use service auto-registration, Symfony *does* automatically add aliases for any interfaces that your services implement. That’s a relatively new and undocumented feature :). I hope this clears things up a bit - and yea, check the docs - they are accurate.

pensiero:
Well, if I write this in a service constructor:
```
    /**
     * @param DocumentManager $dm
     */
    public function __construct(DocumentManager $dm)
    {
        $this->dm = $dm;
    }
```

(with `use Doctrine\ODM\MongoDB\DocumentManager`), a `RuntimeException` is triggered:

> Cannot autowire service "App\Track\TrackManager": argument "$dm" of method "__construct()" references class "Doctrine\ODM\MongoDB\DocumentManager" but no such service exists. Try changing the type-hint to "Doctrine\Common\Persistence\ObjectManager" instead.

If I change to `ObjectManager, as the Exception suggests, it returns the ORM EntityManager (as I said, in my project there are both the ORM and ODM).

In Symfony 3.4 it worked perfectly, but in the profiler there was a deprecation notice about this.

weaverryan:
Exactly right. But this is only because the bundle chose to create an autowiring alias for the interface, but not the class. If you added an alias for the class, you could get it to work also (sorry for not providing a code sample, I’m on my phone!)

pensiero:
Maybe I understand it wrong, are you saying that I should do something like
```
use Doctrine\ODM\MongoDB\DocumentManager as MongoDBDocumentManager;
```
?

If so, it's not working, but probably I got it wrong :/

alcaeus:
Note: changes in this pull request haven't been released yet, so unless you're using `dev-master`, you'll need to create the service alias manually.

pensiero:
@alcaeus thanks, that's why it's not working, I added this lines in my services.yml:
```
    Doctrine\ODM\MongoDB\DocumentManager:
        alias: doctrine_mongodb.odm.document_manager
```
and it works ;)

waiting the next release!

weaverryan:
Nice work. THAT was the code block I wanted to post earlier, but couldn’t - nice job :)

pensiero:
@weaverryan @alcaeus @Majkl578 thanks a lot :)

PR #448, opened 2018-01-19
Update index.rst with minimum documentation for Symfony 4 and autowiring

oriole9g:
I have added minimum documentation for guidance if you are using DoctrineMongoDBBundle in a Symfony 4 project like us.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -46,7 +46,7 @@ To install DoctrineMongoDBBundle with Composer just add the following to your
     }
 
 
-If you are planning to use autowiring, you must install version 3.5 or later::
+If you are planning to use `autowiring`, you must install version 3.5 or later::
 
     {
         "require": {
@@ -910,3 +910,4 @@ Learn more from the Cookbook
 .. _`UniqueEntity`: http://symfony.com/doc/current/reference/constraints/UniqueEntity.html
 .. _`store sessions`: http://symfony.com/doc/current/cookbook/doctrine/mongodb_session_storage.html
 .. _`"Using PHP 7" section`: http://docs.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/reference/introduction.html#using-php-7
+.. _`autowiring`: https://symfony.com/doc/current/service_container/autowiring.html
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -135,9 +135,9 @@ the MongoDB ODM across your application:
     If you are using Symfony Flex, you can allow `recipes` in the "contrib" repository to 
     work with this bundle by executing the following command:
 
-.. code-block:: bash
+    .. code-block:: bash
 
-    composer config extra.symfony.allow-contrib true
+        composer config extra.symfony.allow-contrib true
 
 .. note::
 
@@ -344,7 +344,7 @@ Let's walk through this example:
   to MongoDB. In this example, the ``$product`` object has not been persisted yet,
   so the document manager makes a query to MongoDB, which adds a new entry.
 
-If you are using autowiring, you can use type hinting to fetch the doctrine_mongodb.odm.document_manager service::
+If you are using `autowiring`, you can use type hinting to fetch the ``doctrine_mongodb.odm.document_manager`` service::
 
 .. code-block:: php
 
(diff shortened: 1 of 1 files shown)


malarzm:
@jmikola can you take a look? :)

oriole9g:
I think that I have solved all comments from @jmikola 

The only thing I have not linked is the allow-contrib option for recipes, doesn't seem to have official documentation, only this blog post: http://fabien.potencier.org/symfony4-contributing-recipes.html

oriole9g:
I have linked recipes to Fabien blog post, at this moment this is the best documentation that exists.

Resources/doc/index.rst
```diff
@@ -122,6 +130,15 @@ the MongoDB ODM across your application:
...
                 auto_mapping: true
 
+.. note::
+
+    If you are using Symfony Flex, you can allow `recipes` in the "contrib" repository to 
+    work with this bundle by executing the following command:
+
+.. code-block:: bash
+
+    composer config extra.symfony.allow-contrib true
```
jmikola: Question: how does `.. code-block` work with a `.. note`? In context, I imagine this code block should reside within the note, but I expect it will actually render outside.

Based on [this RST preview](http://rst.ninjs.org/?theme=nature#Li4gbm90ZTo6CgogICAgSWYgeW91IGFyZSB1c2luZyBTeW1mb255IEZsZXgsIHlvdSBjYW4gYWxsb3cgYHJlY2lwZXNgIGluIHRoZSAiY29udHJpYiIgcmVwb3NpdG9yeSB0byAKICAgIHdvcmsgd2l0aCB0aGlzIGJ1bmRsZSBieSBleGVjdXRpbmcgdGhlIGZvbGxvd2luZyBjb21tYW5kOgoKICAgIC4uIGNvZGUtYmxvY2s6OiBiYXNoCgogICAgICBjb21wb3NlciBjb25maWcgZXh0cmEuc3ltZm9ueS5hbGxvdy1jb250cmliIHRydWU=) it looks like we can resolve this by aligning the indent of `.. code-block:: bash` with the paragraph above ("If you are using Symfony Flex..."). The `composer config` line itself will then need to be indented further.

Resources/doc/index.rst
```diff
@@ -327,6 +344,33 @@ Let's walk through this example:
   to MongoDB. In this example, the ``$product`` object has not been persisted yet,
   so the document manager makes a query to MongoDB, which adds a new entry.
 
+If you are using autowiring, you can use type hinting to fetch the doctrine_mongodb.odm.document_manager service::
```
jmikola: Should `doctrine_mongodb.odm.document_manager` get monospace treatment with double backticks?

jmikola: Since `autowiring` is already defined as a link, would it make sense to link this instance of the word as well (as you did on line 49 in this diff)?

ELepolt:
Hey all. I'm very new to this project, so I'm not sure how much I'm missing from other pieces of documentation, but I've attempted to follow the documentation provided and I'm still getting some errors.

The first and most glaring to me is this following line in the generating a bundle
`php bin/console generate:bundle --namespace=Acme/StoreBundle`

Besides [Symfony encouraging users to not use bundles](https://symfony.com/doc/current/bundles.html), that generate:bundle command is through [Sensio](https://github.com/sensiolabs/SensioGeneratorBundle), and they are saying they don't support the bundle-less directory structure.

After creating my own bundle from Symfony's manual directions, I did notice that I needed to add `App\` into my namspace, otherwise I receive the following error: 
`The autoloader expected class "App\Acme\StoreBundle\Document\Product" to be defined in the file [path/to/Product.php]. The file was found but the class was not in it, the class name or namespace probably has a typo.`

Once we figured that out, we were successful in using `doctrine:mongodb:mapping:info`

Finally, when we attempt to run `doctrine:mongodb:generate:documents AcmeStoreBundle` we are giving the following error:
`Can't find base path for bundle (path "App/Acme/StoreBundle", destination [project_name/src/Acme/StoreBundle]`

Which definitely exists.

Again. Not sure if this should be a new issue, or if it's ok to post here. Thank you for all your hard work. I'll be continuing effort to find the correct changes tomorrow.

malarzm:
@ELepolt this seems like a usage question, please post it to http://stackoverflow.com using `symfony` and `doctrine-odm` tags.

ELepolt:
@malarzm, sorry. I didn't mean to necessarily ask help here, but the comments I made will effect every user who attempts to follow this pull request for Symfony 4.

malarzm:
Sorry then, let's move the discussion to a new issue then as generating code is not connected with scope of the PR :)

PR #451, opened 2018-02-28
[WIP] Use doctrine/mongodb-odm 2.0

rubenrua:
*WORK IN PROGRESS*

Use the new version of `doctrine/mongodb-odm` based on `ext-mongodb`


alcaeus:
@rubenrua thank you for the contribution. Just a heads up, since MongoDB ODM 2.0 is nowhere near finalized at this time, we haven't yet started working on this bundle. One reason for it is that at the current stage, we don't recommend using `dev-master` of MongoDB ODM unless you're willing to risk complete data loss, hard failures and BC breaks while we keep working on it.

We'll revisit this bundle once we're in an alpha or beta stage for ODM itself, at which time we'll have lots of cleanup to do here as well (strict typing, proper coding standards, etc.). I'll leave this pull request open, just be aware that it's not going to be merged in the near future.

To reiterate for anyone else reading this: MongoDB ODM 2.0 **should not be used** at this time: it's feature-incomplete, big breaking changes are still to come and it has not been properly tested. We'll let everyone know once we've moved to an alpha or beta phase.

rubenrua:
@alcaeus Thank you for all your work in ODM 2.0. We are looking forward to use in it.

Let me know if you need help with other issues/task related with ODM 2.0.

rubenrua:
FYI: symfony/doctrine-bridge 4.0.2 drops `memcache` type. Where is the Backward Compatibility Promise?

Memcached type is used instead.

See: https://github.com/symfony/doctrine-bridge/commit/d14c17af9290634eb1588b75ed64a64ee286ba3e

jmikola:
> FYI: symfony/doctrine-bridge 4.0.2 drops memcache type. Where is the Backward Compatibility Promise?

@rubenrua: I'd suggest opening an issue on the Symfony repository to ask about that (and linking to https://github.com/symfony/doctrine-bridge/commit/d14c17af9290634eb1588b75ed64a64ee286ba3e for context). I see that you commented on the commit itself, but an issue may receive more visibility.

AFAIK, that package is not maintained by any of the Doctrine developers, and it's certainly not related to the ODM bundle.

rubenrua:
From https://github.com/symfony/symfony/pull/25409
> There is no "memcache" extension for PHP 7. Only memcached works now.

\<irony\>Looks like a Backward Compatibility Break in PHP \<\/irony\>

As the new version of the bundle only support PHP7 the last commit f7aad2b9845cf18f2a140eac11c0263356f4a340 is valid.

alcaeus:
> Where is the Backward Compatibility Promise?

Let's say you started using Symfony 4.0.0 and realized that your `memcache` extension no longer lives: you deal with the BC break while updating to Symfony 4.0. Now, if in the meantime there was a way to get the `memcache` extension back, I'd agree with calling this a BC break, but since nobody could've been using it, it's okay to drop the class in a minor release, even when it's strictly a BC break.

Regardless of the above, switching to `memcached` in this bundle is fine. We'll want to release these changes as a new major version anyways and will also release a 3.5 (probably) release containing tons of deprecations to prepare for the new bundle version.

alcaeus:
Closing in favour of #475 - there are many more changes necessary to get the bundle to work. I'll finish this up in preparation of an alpha release of ODM 2.0. Thanks @rubenrua!

Issue #453, opened 2018-03-19
The authMechanism 'X509' does not exist


chrispindjako:
Hi, I'm using symfony 3.2 with MongoDB 3.4 and in my DoctrineMongoDBBundle 3.4.1 configuration file 
the parameter authMechanism is present with value X509. My problem is that when I try to connect to my Mongo Database i received  this error:  {"code":500,"message":"The authMechanism 'X509' does not exist."}

My DoctrineMongoDBBundle configuration file is below:

![sans titre](https://user-images.githubusercontent.com/16692562/37616435-cbd703ce-2ba7-11e8-93f1-ec4e8344d07e.png)

And my associate servIce to this config file is :

![sans titre](https://user-images.githubusercontent.com/16692562/37616568-489e45de-2ba8-11e8-8b73-e12c90cfba45.png)
Can you help me please to fix this isssue?

jmikola:
> `{"code":500,"message":"The authMechanism 'X509' does not exist."}`

While this may be the response from your web application, we'll need to see the actual PHP exception from which this error originated (its class name, error message, and error code). Additionally, a condensed stack trace going back to the function that threw the exception will be helpful.

Separate from that, details about your MongoDB environment (e.g. [`buildInfo`](https://docs.mongodb.com/manual/reference/command/buildInfo/) command output) and PHP environment (e.g. [`phpinfo()`](http://php.net/manual/en/function.phpinfo.php) output) will be helpful.

chrispindjako:
Here are the informations you asked me:

     -**symfony dev.log**:
![symfony_log_dev](https://user-images.githubusercontent.com/16692562/37662872-f02d0054-2c4f-11e8-9ad2-9ba7dcb9109d.png)

     -**mongodb buildInfo**
![mongodb_info](https://user-images.githubusercontent.com/16692562/37662941-141b6ae6-2c50-11e8-9c4d-1141205ac7ee.png)

     -**php info**:
[phpinfo.txt](https://github.com/doctrine/DoctrineMongoDBBundle/files/1829892/phpinfo.txt)

_thank you in advance_

jmikola:
It looks like that exception message originates in the [`mongo_store_option()`] function (https://github.com/mongodb/mongo-php-driver-legacy/blob/1.6.16/mcon/parse.c#L403) in the legacy "mongo" extension.

[...]

PR #454, opened 2018-03-21
Replace authMechanism option value 'X509' by the proper value

chrispindjako:
'MONGODB-X509'


malarzm:
@chrispindjako thanks a bunch for the PR! I'll try to take a look at failing tests today (these are unrelated to your changes) and tag 3.4.2 then.

jmikola:
Ah, I didn't realize that the bundle was suggesting/requiring an `X509` value for `authMechanism` when I was tending to #453 and wrongfully assumed it was just a documentation issue on PHP.net. Granted, I suppose that's how the mistake in the bundle was introduced.

Thanks, @chrispindjako!

chrispindjako:
It's my pleasure.

PR #458, opened 2018-04-12
Fix type guesser. Ignore field MappingException.

BoShurik:
Fix errors like "No mapping found for field 'plainPassword' in class 'App\Document\Administrator'."

### DIFF
--- a/Form/DoctrineMongoDBTypeGuesser.php
+++ b/Form/DoctrineMongoDBTypeGuesser.php
@@ -17,7 +17,6 @@ use Symfony\Component\Form\FormTypeGuesserInterface;
 use Symfony\Component\Form\Guess\Guess;
 use Symfony\Component\Form\Guess\TypeGuess;
 use Symfony\Component\Form\Guess\ValueGuess;
-use Symfony\Component\HttpKernel\Kernel;
 
 /**
  * Tries to guess form types according to ODM mappings
@@ -44,10 +43,14 @@ class DoctrineMongoDBTypeGuesser implements FormTypeGuesserInterface
     public function guessType($class, $property)
     {
         if (!$ret = $this->getMetadata($class)) {
-            return new TypeGuess($this->typeFQCN ? TextType::class : 'text', [], Guess::LOW_CONFIDENCE);
+            return;
         }
 
         list($metadata, $name) = $ret;
+
+        if (! $metadata->hasField($property)) {
+            return;
+        }
 
         if ($metadata->hasAssociation($property)) {
             $multiple = $metadata->isCollectionValuedAssociation($property);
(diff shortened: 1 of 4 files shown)


Form/DoctrineMongoDBTypeGuesser.php
```diff
@@ -49,6 +49,10 @@ public function guessType($class, $property)
 
         list($metadata, $name) = $ret;
 
+        if (!$metadata->hasField($property)) {
```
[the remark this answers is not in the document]
BoShurik: @malarzm Done. But in code there are only 3 places with such style. And 22 without spaces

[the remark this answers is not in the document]
malarzm: Ah so the bundle did not receive CS fixes yet, thank you :)

Form/DoctrineMongoDBTypeGuesser.php
```diff
@@ -49,6 +49,10 @@ public function guessType($class, $property)
 
         list($metadata, $name) = $ret;
 
+        if (! $metadata->hasField($property)) {
+            return new TypeGuess($this->typeFQCN ? TextType::class : 'text', [], Guess::LOW_CONFIDENCE);
```
malarzm: Now I saw the test I'm wondering whether guessing `TextType` (even with low confidence) is correct. From what I saw in https://symfony.com/doc/current/form/type_guesser.html it's possible to just `return;` if we can't (or shouldn't) make a guess at all. Do you think it'd be more appropiate?

BoShurik: I copy behaviour from a [symfony/doctrine-bridge](https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/Doctrine/Form/DoctrineOrmTypeGuesser.php#L88). Looks like in the first versions type `guessType` returns only `TypeGuess` instance (based on phpDoc). I think now we can return nulls

malarzm: Also if you could squash commits later that'd be awesome :)

BoShurik: Done

BoShurik:
@malarzm ping! Is there any other things that prevent merging?

malarzm:
Sorry, had this email starred but did not get back to it in time. Looks good to me! @alcaeus are we couting this as a bugfix and releasing 3.4.3 or will this wait for 3.5.0?

malarzm:
FWIW I'm fine with relasing 3.4.3 with this fix only as we do not get that many issues to fix.

alcaeus:
Bugfix it is.

PR #467, opened 2018-06-19
Change the IntegerNode port configuration node to ScalarNode…

### LINKED ISSUE #466: Bundle doesn't support port configuration by environment variable
When configuring the metadata_cache_driver section to use environment variables for the port configuration

```yaml
doctrine_mongodb:
    document_managers:
        default:
            metadata_cache_driver:
                type: memcached
                class: Doctrine\Common\Cache\MemcachedCache
                instance_class: Memcached
                host: '%env(DOCTRINE_MONGODB_CACHE_HOST)%'
                port: '%env(int:DOCTRINE_MONGODB_CACHE_PORT)%'
```
It'll fail building the container with the exception
```
In IntegerNode.php line 29:

  Invalid type for path "doctrine_mongodb.document_managers.default.metadata_cache_driver.port". Expected int, but got string.
```

This makes sense because the port configuration should be an integer as it is configured as an IntegerNode. This is therefore not going to work because the placeholder value is being a string.

I would propose to change the port node to be of type `ScalarNode`, that would also be consistent with the DBAL configuration section which accepts the port to be scalar as well which you can check here: https://github.com/doctrine/DoctrineBundle/blob/master/DependencyInjection/Configuration.php#L192

netiul:
Change the IntegerNode port configuration node to ScalarNode to accept placeholder environment variables

Fixes #466

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -176,7 +176,7 @@ class Configuration implements ConfigurationInterface
                                     ->scalarNode('type')->defaultValue('array')->end()
                                     ->scalarNode('class')->end()
                                     ->scalarNode('host')->end()
-                                    ->integerNode('port')->end()
+                                    ->scalarNode('port')->end()
                                     ->scalarNode('instance_class')->end()
                                     ->scalarNode('id')->end()
                                     ->scalarNode('namespace')->end()


alcaeus:
Closing as per discussion in #466: Symfony should support this. I'm not willing to change the port number to any scalar type - it's an int by definition and it should remain that way.

netiul:
Understand that 👍

PR #472, opened 2018-09-12
Website docs

jwage:


.doctrine-project.json
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "Doctrine MongoDB Bundle",
+    "shortName": "DoctrineMongoDBBundle",
+    "slug": "doctrine-mongodb-bundle",
+    "versions": [
+        {
+            "name": "4.0",
```
alcaeus: `master` is still `3.5-dev` and will remain so until we tag 3.5.0, so this should be updated.

jwage: Updated.

PR #473, opened 2018-09-17
Add support for service repositories

alcaeus:
This is a copy of the work @weaverryan has done for DoctrineBundle in https://github.com/doctrine/DoctrineBundle/pull/727 and brings service repositories to the ODM bundle. An example (also shamelessly copy/pasted from the original PR):

```php
/**
 * @ODM\Document(repositoryClass=CoolStuffRepository::class)
 */
class CoolStuff
{

}
```

Your repository class will (with default config) auto-registered as a service. So, without touching any config, you create the class and it just works:

```php
class CoolStuffRepository extends ServiceDocumentRepository /* this class extends DocumentRepository, implements ServiceDocumentRepositoryInterface */
{
    public function __construct(RegistryInterface $registry)
    {
        parent::__construct($registry, CoolStuff::class);
    }

    // your custom methods
}
```

The feature is completely optional - if the repository factory encounters a repository that does not implement the `ServiceDocumentRepositoryInterface` marker, it will create the repository as it used to. The `ServiceDocumentRepository` class is a helper that doesn't have to be used (it's recommended you use it).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,6 +12,7 @@
     ],
     "require": {
         "php": "^5.6 || ^7.0",
+        "doctrine/annotations": "^1.2",
         "doctrine/mongodb": "^1.4",
         "doctrine/mongodb-odm": "^1.1",
         "psr/log": "^1.0",


DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -77,6 +79,13 @@ public function load(array $configs, ContainerBuilder $container)
 
         // BC Aliases for Document Manager
         $container->setAlias('doctrine.odm.mongodb.document_manager', new Alias('doctrine_mongodb.odm.document_manager'));
+
+
+        $container->registerForAutoconfiguration(ServiceDocumentRepositoryInterface::class)
+            ->addTag(ServiceRepositoryCompilerPass::REPOSITORY_SERVICE_TAG);
+
+        $container->getDefinition('doctrine_mongodb.odm.container_repository_factory')
```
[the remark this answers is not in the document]
weaverryan: Ooof. I can't remember.

Here is my guess. This feature only works with 3.3 and higher... because you need the `ServiceLocatorTagPass`. I *think* (not sure) that I was trying to avoid a nasty error in 3.2. Specifically, if you set the `ServiceLocator` into your XML file, that would fail on 3.2 - as that class doesn't exist. However, I'm not sure why I didn't just leave that argument blank. 

Look at the original commit : https://github.com/doctrine/DoctrineBundle/commit/4efcb135344e55403a0fd193c23a8c7249cfd7d0#diff-8f3578ddca2c59937b37c3331b2ab336R388

I set the first argument to null in XML (https://github.com/doctrine/DoctrineBundle/commit/4efcb135344e55403a0fd193c23a8c7249cfd7d0#diff-8f3578ddca2c59937b37c3331b2ab336R388) to avoid errors on 3.2 I believe. But... I'm not sure why I set this to an empty ServiceLocator in the extension class. I'm *guessing* that there was something that was *forcing* me to have this properly set. Perhaps a type-check on my `null` versus the type-hint on that argument. That would make sense :). 

If you're not supporting Symfony 3.2... I'm not sure any of this is necessary. Here's an example from core - it looks super simple: https://github.com/symfony/symfony/blob/f834c9262b411aa5793fcea23694e3ad3b5acbb4/src/Symfony/Bundle/TwigBundle/DependencyInjection/Compiler/RuntimeLoaderPass.php#L37 and the definition: https://github.com/symfony/symfony/blob/f834c9262b411aa5793fcea23694e3ad3b5acbb4/src/Symfony/Bundle/TwigBundle/Resources/config/twig.xml#L159

Try that simpler route - see if there are any issues.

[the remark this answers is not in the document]
alcaeus: I’ll give that a shot, thank you for your help!

Repository/ContainerRepositoryFactory.php
```diff
@@ -0,0 +1,78 @@
...
+    private function getOrCreateRepository(DocumentManager $documentManager, ClassMetadata $metadata)
+    {
+        $repositoryHash = $metadata->getName() . spl_object_hash($documentManager);
+        if (isset($this->managedRepositories[$repositoryHash])) {
+            return $this->managedRepositories[$repositoryHash];
+        }
+
+        $repositoryClassName = $metadata->customRepositoryClassName ?: $documentManager->getConfiguration()->getDefaultRepositoryClassName();
+
+        return $this->managedRepositories[$repositoryHash] = new $repositoryClassName($documentManager, $documentManager->getUnitOfWork(), $metadata);
```
malarzm: Shouldn't this defer repository creation to `DocumentManager` (which would then defer creation to repository factory)? Right now custom repositories with additional dependencies will be broken if I'm not mistaken

Repository/ServiceDocumentRepository.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace Doctrine\Bundle\MongoDBBundle\Repository;
+
+use Doctrine\Common\Persistence\ManagerRegistry;
+use Doctrine\ODM\MongoDB\DocumentManager;
+use Doctrine\ODM\MongoDB\DocumentRepository;
+
+/**
+ * Optional EntityRepository base class with a simplified constructor (for autowiring).
```
malarzm: DocumentRepository :)

Repository/ServiceDocumentRepository.php
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * To use in your class, inject the "registry" service and call
+ * the parent constructor. For example:
+ *
+ * class YourEntityRepository extends ServiceEntityRepository
```
malarzm: Here as well

Repository/ServiceDocumentRepository.php
```diff
@@ -0,0 +1,35 @@
...
+ * {
+ *     public function __construct(RegistryInterface $registry)
+ *     {
+ *         parent::__construct($registry, YourEntity::class);
```
malarzm: `YourDocument::class` ;)

Repository/ServiceDocumentRepository.php
```diff
@@ -0,0 +1,35 @@
...
+ *     }
+ * }
+ */
+class ServiceDocumentRepository extends DocumentRepository implements ServiceDocumentRepositoryInterface
+{
+    /**
+     * @param string $entityClass The class name of the entity this repository manages
+     */
+    public function __construct(ManagerRegistry $registry, $entityClass)
```
malarzm: Parameter should be named `$documentClass`

Repository/ContainerRepositoryFactory.php
alcaeus: I didn't want to wrap the default factory, but could certainly do so. Either way, custom repositories with additional dependencies **are** broken, which is why this PR will allow you to declare them as services to be fetched by this new factory.

DependencyInjection/DoctrineMongoDBExtension.php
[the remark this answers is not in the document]
alcaeus: Dropping `null` as well as the `replaceArgument` call in the extension works. Not sure about Symfony < 3.4, but we don't use those here.

Repository/ContainerRepositoryFactory.php
malarzm: Ahhh all right, I think I've wrapped my head around this. This class could extend [`AbstractRepositoryFactory`](https://github.com/doctrine/mongodb-odm/blob/1.3.x/lib/Doctrine/ODM/MongoDB/Repository/AbstractRepositoryFactory.php), this way you'll get rid of managing instances here. WDYT?

alcaeus: I'm not sure we want this. The original implementation always fetches the repository service from the container without it caching it locally. Only repositories that have to be instantiated are cached. That's the reason why I didn't use the abstract factory from the start since I was trying to keep this behaviour intact. It could make a difference when a service isn't shared, e.g. a new instance is returned when getting it from the container. AFAIK that also holds true when using `ServiceLocatorTagPass` to register a service locator with a bunch of services.

PR #475, opened 2018-09-24
Prepare bundle for ODM 2.0

alcaeus:
* [x] Drop PHP < 7.2
* [x] Require development versions of ODM 2.0 (will change as we move into alpha and beta stages)
* [x] Drop YAML driver
* [x] Update configuration to instantiate new connection classes
* [x] Make service repositories compatible with GridFS repositories
* [x] Use PHPUnit 7.3
* [x] Update data collectors for `ext-mongodb` (needs 
https://github.com/doctrine/mongodb-odm/pull/1866)
* [x] Update query logger for `ext-mongodb`


APM/PSRCommandLogger.php
```diff
@@ -0,0 +1,65 @@
...
+use function json_encode;
+use function MongoDB\Driver\Monitoring\addSubscriber;
+use function MongoDB\Driver\Monitoring\removeSubscriber;
+
+final class PSRCommandLogger implements CommandSubscriber
+{
+    /** @var bool */
+    private $registered = false;
+
+    /** @var LoggerInterface */
```
malarzm: According to ctor this can be nullable

alcaeus: Ah, good catch. Thanks!

PR #482, opened 2018-10-02
Drop document generation commands

### LINKED ISSUE #480: Drop document generation commands
See https://github.com/doctrine/DoctrineMongoDBBundle/issues/479#issuecomment-426149593: `DocumentGenerator` has been dropped from ODM, so the respective commands need to be dropped from the bundle in time for Beta 1.

alcaeus:
Fixes #480. I'll create a separate PR to replace the other generation commands with those from MakerBundle, just need to familiarise myself with it a little more.


alcaeus:
@malarzm rebased on top of master and added notice to `UPGRADE-4.0` document.

PR #484, opened 2018-10-03
Apply doctrine/coding-standard 5.0

### LINKED ISSUE #476: Apply doctrine/coding-standard 5.0

caciobanu:
Fixes #476


phpcs.xml.dist
```diff
@@ -0,0 +1,56 @@
...
+    <arg name="parallel" value="80"/>
+    <arg name="cache" value=".phpcs-cache"/>
+    <arg name="colors" />
+
+    <!-- Ignore warnings and show progress of the run -->
+    <arg value="np"/>
+
+    <file>.</file>
+
+    <exclude-pattern>vendor/*</exclude-pattern>
```
carusogabriel: This is excluded by default, we can remove this line

phpcs.xml.dist
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0"?>
+<ruleset>
+    <arg name="basepath" value="."/>
+    <arg name="extensions" value="php"/>
+    <arg name="parallel" value="80"/>
+    <arg name="cache" value=".phpcs-cache"/>
+    <arg name="colors" />
```
carusogabriel:
```diff
-<arg name="colors" />
+<arg name="colors"/>
```

phpcs.xml.dist
```diff
@@ -0,0 +1,56 @@
...
+
+    <exclude-pattern>vendor/*</exclude-pattern>
+
+    <rule ref="Doctrine">
+        <!-- Traversable type hints often end up as mixed[], so we skip them for now -->
+        <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingTraversablePropertyTypeHintSpecification" />
+        <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingTraversableParameterTypeHintSpecification" />
+        <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingTraversableReturnTypeHintSpecification" />
+
+        <!-- Will cause BC breaks to method signatures - disabled for now -->
```
carusogabriel: If we added `declare(strict_types=1)` I see no reason why we can't change the signatures of the methods, as both are BC breaks.

phpcs.xml.dist
```diff
@@ -0,0 +1,56 @@
...
+    <arg name="basepath" value="."/>
+    <arg name="extensions" value="php"/>
+    <arg name="parallel" value="80"/>
+    <arg name="cache" value=".phpcs-cache"/>
+    <arg name="colors" />
+
+    <!-- Ignore warnings and show progress of the run -->
+    <arg value="np"/>
+
+    <file>.</file>
```
carusogabriel: I'd rather list all folders and files we are tracking with PHPCS

phpcs.xml.dist
alcaeus: The root version of the standard was copied from ODM itself where we didn't want to add more BC breaks than absolutely necessary. We can revisit this here for 4.0.

phpcs.xml.dist
alcaeus: Go for it 👍

alcaeus:
@carusogabriel regarding composer.lock, we consciously took the decision to not include this here or in mongodb-odm, knowing full well that inadvertent BC breaks or new sniffs in the phpcs ruleset could break the build. I don't see a reason to change that at this stage.

PR #487, opened 2018-10-05
Patch for SF4 support

Romaxx:
SF4 patch for single App Namespace


malarzm:
@Romaxx thank you for the PR! Please target branch 3.5 as the command you're changing will be removed in 4.0 (which will be tagged from the master branch). Also there are multiple code-style issues, I'm not sure whether it'll be possible to fix them in 3.5 branch with `phpcs`, if not I'll point them out once the build passes.

alcaeus:
This is a new feature, as such it won't be part of 3.5. Since we're dropping the command in the next release anyways, I don't think it's worth the effort to fix this.

Issue #488, opened 2018-10-07
CommandDataCollector 


blixit:
I'm trying to use the DoctrineMongoDBBundle  with symfony4.
After the bundle installation, 
`composer require doctrine/mongodb-odm-bundle`
I got this error: 

Compile Error: Declaration of `Doctrine\Bundle\MongoDBBundle\DataCollector\CommandDataCollector::collect(Symfony\Component\HttpFoundation\Request $request, Symfony\Component\HttpFoundation\Response $response, ?Throwable $exception = NULL) `
must be compatible with `Symfony\Component\HttpKernel\DataCollector\DataCollectorInterface::collect(Symfony\Component\HttpFoundation\Request $request, Symfony\Component\HttpFoundation\Response $response, ?Exception $exception = NULL)`

This error can be fixed by changing **Throwable** by **Exception**.

malarzm:
Indeed, the [signature contains `Exception` for 6 years now](https://github.com/symfony/symfony/blame/cbe289517470eeea27162fd2d523eb29c95f775f/src/Symfony/Component/HttpKernel/DataCollector/DataCollectorInterface.php#L28), mind sending us a PR? :)

blixit:
> Indeed, the [signature contains `Exception` for 6 years now](https://github.com/symfony/symfony/blame/cbe289517470eeea27162fd2d523eb29c95f775f/src/Symfony/Component/HttpKernel/DataCollector/DataCollectorInterface.php#L28), mind sending us a PR? :)

Of course. let me fork the project.

alcaeus:
We need to check some more classes - #484 added doctrine/coding-standard which requires using `Throwable` instead of `Exception`.

alcaeus:
Yes please! 👍

jcaillot:
Symfony\Component\HttpKernel\DataCollector\DataCollectorInterface interface changed (again) with the 5.0 version !

we have got :

Symfony 4.2

` public function collect(Request $request, Response $response, \Exception $exception = null);
 `
https://github.com/symfony/symfony/blob/4.2/src/Symfony/Component/HttpKernel/DataCollector/DataCollectorInterface.php


Symfony 4.3 and 4.4

`public function collect(Request $request, Response $response/*, \Throwable $exception = null*/);`

https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/HttpKernel/DataCollector/DataCollectorInterface.php


Symfony 5.0

[...]

PR #489, opened 2018-10-08
Changed exception type

blixit:
Changed exception interface 'Throwable' by the explicit class '\Exception' in order to fix https://github.com/doctrine/DoctrineMongoDBBundle/issues/488

### DIFF
--- a/DataCollector/CommandDataCollector.php
+++ b/DataCollector/CommandDataCollector.php
@@ -6,6 +6,7 @@ namespace Doctrine\Bundle\MongoDBBundle\DataCollector;
 
 use Doctrine\ODM\MongoDB\APM\Command;
 use Doctrine\ODM\MongoDB\APM\CommandLogger;
+use Exception;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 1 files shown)


DataCollector/CommandDataCollector.php
```diff
@@ -24,7 +23,7 @@ public function __construct(CommandLogger $commandLogger)
         $this->commandLogger = $commandLogger;
     }
 
-    public function collect(Request $request, Response $response, ?Throwable $exception = null)
+    public function collect(Request $request, Response $response, ?\Exception $exception = null)
```
alcaeus: Needs a `phpcs:ignore` statement in the line above:
```
// phpcs:ignore SlevomatCodingStandard.Exceptions.ReferenceThrowableOnly
```

Also, `Exception` shouldn't be referenced by global name but rather imported above (as was done for `Throwable`). You can check whether your changes worked by running `composer install` and then `vendor/bin/phpcs`.

blixit:
Ok. This week I am busy, I will fix it this end of week, probably saturday.

alcaeus:
Thanks @blixit!

PR #491, opened 2018-10-12
Update form.rst

polishExperiment:


alcaeus:
Thanks for the contribution, @polishExperiment. In the future, please use more meaningful commit and pull request messages instead of relying on the default messages GitHub provides. Thanks.

PR #493, opened 2018-10-15
Update form.rst

polishExperiment:
form_enctype is deprecated


alcaeus:
As I've previously stated, please use meaningful commit messages and pull request **titles**. If I see a PR with "Update form.rst" in my notifications chances are I'll just ignore it. Thanks!

PR #502, opened 2018-11-16
Fix wrong capitalization of typeMap option

### LINKED ISSUE #494: Invalid array key
On [DoctrineMongoDBExtension::normalizeDriverOptions:319](https://github.com/doctrine/DoctrineMongoDBBundle/blob/master/DependencyInjection/DoctrineMongoDBExtension.php#L319) we have `typemap` (as well as on all tests) but on `ext-mongodb` `Client` class constructor the key is `typeMap`. This generate various problems with `array` type hints on `Doctrine\ODM\MongoDB\UnitOfWork:getOrCreateDocument` for example which expect an array but get a `MongoDB\Model\BSONArray` (which is the default and is not an array).

A var_dump of generated option:
```
Array
(
    [typemap] => Array
        (
            [root] => array
            [document] => array
        )
    [typeMap] => Array
        (
            [array] => MongoDB\Model\BSONArray
            [document] => MongoDB\Model\BSONDocument
            [root] => MongoDB\Model\BSONDocument
        )
)
```
Change row 319 to:
```
$driverOptions['typeMap'] = ['root' => 'array', 'document' => 'array'];
```
and the problem is solved.

alcaeus:
Fixes #494

### DIFF
--- a/DependencyInjection/DoctrineMongoDBExtension.php
+++ b/DependencyInjection/DoctrineMongoDBExtension.php
@@ -316,7 +316,7 @@ class DoctrineMongoDBExtension extends AbstractDoctrineExtension
     private function normalizeDriverOptions(array $connection)
     {
         $driverOptions            = $connection['driverOptions'] ?? [];
-        $driverOptions['typemap'] = ['root' => 'array', 'document' => 'array'];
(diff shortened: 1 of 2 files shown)


malarzm:
Thanks @alcaeus! ;)

Issue #504, opened 2018-11-29
Call to a member function getUnitOfWork() on null when using ServiceDocumentRepository in Symfony 4


adityarai:
```php
<?php

namespace App\Repository;

use App\Document\HtmlView;
use Doctrine\Bundle\MongoDBBundle\Repository\ServiceDocumentRepository;
use Symfony\Bridge\Doctrine\RegistryInterface;

/**
 * @method HtmlView|null find($id, $lockMode = null, $lockVersion = null)
 * @method HtmlView|null findOneBy(array $criteria, array $orderBy = null)
 * @method HtmlView[]    findAll()
 * @method HtmlView[]    findBy(array $criteria, array $orderBy = null, $limit = null, $offset = null)
 */
class HtmlViewRepository extends ServiceDocumentRepository
{
    public function __construct(RegistryInterface $registry)
    {
        parent::__construct($registry, HtmlView::class);
    }
}
```

alcaeus:
Please provide additional information, e.g. the version of the bundle you are using and a full trace for the exception in question. The information you provided is not sufficient to take a look at this.

adityarai:
Bundle version is 
"alcaeus/mongo-php-adapter": "^1.1",
"doctrine/mongodb-odm-bundle": "^3.5",

controller Action:
`/**
     * @Route("/", name="html_view_index", methods="GET")
     */
    public function index(HtmlViewRepository $htmlViewRepository): Response
    {
        return $this->render('html_view/index.html.twig', ['html_views' => $htmlViewRepository->findAll()]);
    }`

![call to a member function getunitofwork on null 500 internal server error](https://user-images.githubusercontent.com/1521692/49207539-39541800-f3db-11e8-8569-a88f26bd386a.png)

stof:
@adityarai FYI, a much better way to provide the stack trace in a bug report is to switch to the `Stack Trace` tab on that error page, which provides the stack trace in a plain-text format (which can be copy-pasted in the issue)

alcaeus:
That’s a good point - I’ll add a nice exception message to the trait to make spotting these issues easier.

adityarai:
![call to a member function getunitofwork on null 500 internal server error 1](https://user-images.githubusercontent.com/1521692/49215106-06ffe600-f3ee-11e8-870a-474d8fee68d4.png)

adityarai:
Symfony version is 4.1.8 

In the service.yaml:

services:
[...]

PR #505, opened 2018-11-30
Better error message when missing document mapping

alcaeus:
ODM port of https://github.com/doctrine/DoctrineBundle/pull/825. This should help with issues like #504.


Tests/ServiceRepositoryTest.php
```diff
@@ -14,38 +14,38 @@
...
+    private $container;
+
     protected function setUp()
     {
         parent::setUp();
 
-        if (class_exists(DocumentManager::class)) {
-            return;
+        if (! class_exists(DocumentManager::class)) {
+            $this->markTestSkipped('Doctrine MongoDB ODM is not available.');
```
malarzm: I realize this was here before, but why would this happen?

alcaeus: This was copied over from ORM - I'll drop it 👍

PR #506, opened 2018-12-03
Fix more deprecations with Symfony 4.2

alcaeus:
This drops usage of `ContainerAwareCommand` in our commands, replacing it with `ContainerAwareInterface` instead. It also fixes build issues due to missing kernel parameters when testing against Symfony 4.2.

### DIFF
--- a/Command/DoctrineODMCommand.php
+++ b/Command/DoctrineODMCommand.php
@@ -3,11 +3,15 @@
 
 namespace Doctrine\Bundle\MongoDBBundle\Command;
 
+use Doctrine\Bundle\MongoDBBundle\ManagerRegistry;
 use Doctrine\ODM\MongoDB\Tools\DisconnectedClassMetadataFactory;
 use Doctrine\ODM\MongoDB\Tools\DocumentGenerator;
 use Doctrine\ODM\MongoDB\Tools\Console\Helper\DocumentManagerHelper;
-use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 
 /**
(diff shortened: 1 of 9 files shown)


Command/DoctrineODMCommand.php
```diff
@@ -6,20 +6,24 @@
...
  */
-abstract class DoctrineODMCommand extends ContainerAwareCommand
+abstract class DoctrineODMCommand extends Command implements ContainerAwareInterface
 {
+    use ContainerAwareTrait;
+
     public static function setApplicationDocumentManager(Application $application, $dmName)
     {
-        $dm = $application->getKernel()->getContainer()->get('doctrine_mongodb')->getManager($dmName);
+        $dm = $application->getKernel()->container->get('doctrine_mongodb')->getManager($dmName);
```
stof: this looks wrong to me. The property is not public.

Command/DoctrineODMCommand.php
```diff
@@ -38,7 +42,7 @@ protected function getDocumentGenerator()
 
     protected function getDoctrineDocumentManagers()
     {
-        return $this->getContainer()->get('doctrine_mongodb')->getManagers();
+        return $this->container()->get('doctrine_mongodb')->getManagers();
```
stof: the braces look wrong here

alcaeus:
Thanks @stof. It seems my find/replace game was not on point this morning. I'll fix the issues.

alcaeus:
Review issues fixed. I've also added a `getContainer` method to preserve BC. The only (tiny) BC break contained is that the class no longer extends `ContainerAwareCommand`, but any code checking for `ContainerAwareInterface` will still work as expected, although with a deprecation message when calling `getContainer`.

Command/DoctrineODMCommand.php
```diff
@@ -6,17 +6,38 @@
...
 {
+    use ContainerAwareTrait;
+
+    /**
+     * @return ContainerInterface
+     */
+    protected function getContainer()
+    {
+        @trigger_error(
+            sprintf('The %s method is deprecated and should not be used. Please wire your dependencies explicitly.', __METHOD__),
```
malarzm: Is "wire" synonym for "inject" now in this context?

alcaeus:
@malarzm dropped the deprecations: we may only deprecate functionality in minor releases and I don't want to tag 3.6 just yet. So, this will fix the deprecation warnings from Symfony about using `ContainerAwareCommand` while not deprecating instantiating the services without explicit dependencies.

I'll create a separate PR with some more deprecations (including the `retry_error` and `retry_connect` configuration options) which will then be tagged as 3.6 to aid with the migration to 4.0 👍

malarzm:
🎉

khaperets:
Forgot about the command https://github.com/doctrine/DoctrineMongoDBBundle/blob/master/Command/TailCursorDoctrineODMCommand.php

alcaeus:
Thanks for the hint. `DoctrineODMCommand` was already updated, but not in master. There's #513 to fix the deprecation in `TailCursorDoctrineODMCommand`, after which 3.5 will be merged into master for 4.0.0-Beta1.

PR #508, opened 2018-12-03
Fix #507 Symfony 4.2 deprecation

### LINKED ISSUE #507: Symfony 4.2 deprecation : A tree builder without a root node is deprecated since Symfony 4.2 and will not be supported anymore in 5.0.
It seems the bundle currently uses a builder without a root node.

arendjantetteroo:
https://github.com/doctrine/DoctrineMongoDBBundle/issues/507

Not sure how far the BC goes, so if this will be for 4.0 or 3.5?

It seems 4.1 and previous don't have a __construct on the TreeBuilder, so i think you need to have some version check instead?

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -34,8 +34,8 @@ class Configuration implements ConfigurationInterface
      */
     public function getConfigTreeBuilder()
     {
-        $treeBuilder = new TreeBuilder();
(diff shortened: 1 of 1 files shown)


alcaeus:
This is solved in 3.5.1 but hasn't been merged down to master just yet. There's also #506 to solve the `ContainerAwareCommand` deprecation which will be released as 3.5.2. As such, I'm closing this PR - thank you for the contribution.

Issue #75, opened 2012-02-04
Custom Types


henrikbjorn:
I had an idea to create a better flow for handing uploaded files in Symfony2 through a custom type that would mape the file from File to a string and then back again. Unfortunately there is no place to add a custom type like in DoctrineBundle.

Theese are some of the things i have tried.
1. In a bundles build method. This is unfortunately only ran when the Container is build.
2. in a bundles boot method but that is too late because of Doctrine Bundle being enabled before my custom one.

Any further ideas without implementing this directly would be good ;)

hairmare:
Has this been resolved yet?

I was about to dive in and write a compilerpass that finds tagged custom types and then applies them like described on [this blog](http://php-and-symfony.matthiasnoback.nl/2012/09/symfony2-mongodb-odm-creating-custom-types-with-dependencies/).

Would that be a way to do this and would you want me to upstream it ilke that?

dadeky:
Does anyone know when will this issue be resolved?

alcaeus:
I've scheduled this for an upcoming 3.3.0 release, but I can't give you a timeline as to when it will happen.

TheHett:
it would be very convenient to be able to configure custom data types like in the dbal
https://symfony.com/doc/current/doctrine/dbal.html
```
doctrine:
    dbal:
        types:
            custom_first:  App\Type\CustomFirst
            custom_second: App\Type\CustomSecond
```

juliusstoerrle:
bump :) would like to see this coming. Where are you currently registering your types?

juliusstoerrle:
for now I added following to my Kernel::boot() method. Not sure if it's the best place to register types?

`  public function boot()
    {
        parent::boot();
        try {
            $types = $this->getContainer()->getParameter('odm.types');
            foreach ($types as $key => $fqcn) {
                Type::addType($key, $fqcn);
            }
        } catch (\InvalidArgumentException $e) {
            // do nothing if parameter not defined
        }
    }
`

juliusstoerrle:
I would propose adding above code to the boot() method of [DoctrineMongoDBBundle.php](https://github.com/doctrine/DoctrineMongoDBBundle/blob/master/DoctrineMongoDBBundle.php)
With the types listed under the config key "doctrine_mongodb.types".

@alcaeus would that solution suit you? I could prepare a PR.

alcaeus:
Yes, a PR would be a good starting point 👍

[...]

PR #520, opened 2019-01-17
Add registration of custom types in config

juliusstoerrle:
Adds registration of custom types via configuration, similar to Doctrine ORM.

See Issue: #75

Still need to update config checks and test the code, therefore not yet ready for review.

### DIFF
--- a/CacheWarmer/HydratorCacheWarmer.php
+++ b/CacheWarmer/HydratorCacheWarmer.php
@@ -7,6 +7,7 @@ namespace Doctrine\Bundle\MongoDBBundle\CacheWarmer;
 use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\ODM\MongoDB\Configuration;
 use Doctrine\ODM\MongoDB\DocumentManager;
+use Doctrine\ODM\MongoDB\Types\Type;
 use RuntimeException;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpKernel\CacheWarmer\CacheWarmerInterface;
@@ -44,6 +45,9 @@ class HydratorCacheWarmer implements CacheWarmerInterface
 
     public function warmUp($cacheDir)
     {
+        // register custom types during cache warm up to avoid errors while generating hydrator classes
+        $this->registerTypes();
+
         // we need the directory no matter the hydrator cache generation strategy.
         $hydratorCacheDir = $this->container->getParameter('doctrine_mongodb.odm.hydrator_dir');
         if (! file_exists($hydratorCacheDir)) {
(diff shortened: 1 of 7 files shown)


alcaeus:
@juliusstoerrle what's the status here? Do you need help or guidance finishing this PR?

juliusstoerrle:
I got a pretty stable version locally now. I will try to finish everything I'm able to fo myself tomorrow and then we can see. I guess there will be some things that could be improved or need to be added.

juliusstoerrle:
So overall this now seems to work. I initially only registered the types in the `Bundle::boot()`-Method but that resulted in errors during container build time. Only registering during build time obviously doesn't work as that code isn't executed in production. Result of that situation is some code duplication, namely the `registerTypes()` functions in the `Bundle` and the `HydratorCacheWarmer` classes. If someone has a better solution I'm happy to change that.

After some research I added a type element to the xml schema, but as I don't use XML config I did not test it. Is there some easy way to test those changes? (the existing unit tests pass)

I did not yet include any changes to the docs, would you like those in this PR or in a separate one?

juliusstoerrle:
To register a custom type when using **yaml** config:

1.  Add a `types:` key to your `doctrine_mongodb` yaml config
2.  Add your type as `key: value` (`name: FQCN`) pair to the `types` key, like below:
```
doctrine_mongodb:
    ...
    types:
        entity_name: 'App\Infrastructure\Doctrine\ODM\Type\EntityNameType'
        form_description: 'App\Infrastructure\Doctrine\ODM\Type\FormDescriptionType'
```
If you use **XML** config, then add for each type an Element like this to your config:
`<doctrine:type name="foo_bar">Bar\FooBarType</doctrine:type>`

CacheWarmer/HydratorCacheWarmer.php
```diff
@@ -66,4 +70,21 @@ public function warmUp($cacheDir)
             $dm->getHydratorFactory()->generateHydratorClasses($classes);
         }
     }
+
+    private function registerTypes() : void
+    {
+        if (! $this->container->hasParameter('doctrine_mongodb.odm.types')) {
+            // no types defined
+            exit;
```
alcaeus: I'm pretty sure this should be a `return`, not an `exit` ;)
```suggestion
            return;
```

CacheWarmer/HydratorCacheWarmer.php
```diff
@@ -44,6 +45,9 @@ public function isOptional()
 
     public function warmUp($cacheDir)
     {
+        // register custom types during cache warm up to avoid errors while generating hydrator classes
+        $this->registerTypes();
```
alcaeus: Why is the logic from the bundle's `boot` method duplicated here? Shouldn't the bundle be booted, thus types being registered?

juliusstoerrle: I initially only registered the types in the Bundle::boot()-Method but that resulted in errors during container build time. Only registering during build time obviously doesn't work as that code isn't executed in production. Result of that situation is some code duplication, namely the registerTypes() functions in the Bundle and the HydratorCacheWarmer classes. If someone has a better solution I'm happy to change that.

juliusstoerrle: > Shouldn't the bundle be booted, thus types being registered?

Thats what I initally expected aswell.

alcaeus:
Closing in favour of #631.

PR #524, opened 2019-01-19
Replace legacy Doctrine common code

alanpoulain:
Because the dependency to `doctrine/common` is not in the `composer.json`.


alcaeus:
Thanks @alanpoulain!

Issue #509, opened 2018-12-04
Extract fixture loading command to separate bundle


alcaeus:
To better separate concerns and keep track of dependencies, we should extract the `LoadDataFixturesDoctrineODMCommand` to a separate bundle and release it as a separate package (comparable to DoctrineBundle and DoctrineFixturesBundle).

malarzm:
While reason is plausible, I'm not sure extracting a single command to a bundle is the best course of action. 

Offhand, https://github.com/doctrine/DoctrineFixturesBundle is mentioning ODM in the readme, does that bundle support ODM already?

jwage:
Yes, I believe so.

alcaeus:
At the moment, the bundle does not support ODM - I'll check if it can easily be changed to do so. The reason I want to extract it is to have a purpose-built bundle that integrates mongodb-odm and data-fixtures. At this stage, you see the command but can't run it unless you know that you have to require data-fixtures as well.

GromNaN:
I don't think it's useful to extract the code related to fixtures into a separate package, which would require additional maintenance work.

In any case, when a developer wants to add fixtures to their project, they must implement `Doctrine\Common\DataFixtures\FixtureInterface`; they must therefore explicitly require the `doctrine/data-fixtures` package so as not to depend on an indirect dependency.

The `doctrine:mongodb:fixtures:load` command is not available when the `doctrine/data-fixtures` package is not available. https://github.com/doctrine/DoctrineMongoDBBundle/blob/cb65826e2fdcb652e98911b1fc56e1ce736ea2a9/src/DependencyInjection/DoctrineMongoDBExtension.php#L480-L487

PR #529, opened 2019-02-05
First implementation of symfony fixture loader + group support #461

### LINKED ISSUE #461: Fixtures are not loaded as services
According to the [Symfony docs](https://symfony.com/doc/master/bundles/DoctrineFixturesBundle/index.html#accessing-services-from-the-fixtures):
> Your fixtures class is a service, so you can use normal dependency injection

This is currently only true for ORM but not ODM, as this bundle will always instantiate the fixture using `Doctrine\Common\DataFixtures\Loader::createFixture()`.

I would expect that the bundle behaves identical to the ORM part: fixtures that are configured as services are picked up by the CompilerPass and passed to the loader. This loader should be used by the command to trigger fixtures.

A temporary workaround is to implement `ContainerAwareInterface` and use `$this->get(...)` but that is limited to public services.

arendjantetteroo:
A first implementation, basically copy/paste from the fixturesbundle. 

Couple questions:
- Do we want to keep depending on the fixturesBundle or should we copy any files over so we can avoid the fixturesBundle which also installs the orm stuff which if you only use odm you don't need? (for example the FixtureGroupInterface)
- i based it on 3.5.3 as that is what i have locally, it would be nice if this could be included in 3.6, but not sure we can keep it backward compatible? Maybe if we keep fixtures/bundle options and keep their old implementation and if you provide an --autowire option it uses the new one? 

Thoughts?

Fixes #461

### DIFF
--- a/Command/LoadDataFixturesDoctrineODMCommand.php
+++ b/Command/LoadDataFixturesDoctrineODMCommand.php
@@ -3,6 +3,7 @@
 
 namespace Doctrine\Bundle\MongoDBBundle\Command;
 
+use Doctrine\Bundle\MongoDBBundle\Loader\SymfonyFixturesLoaderInterface;
 use Doctrine\Bundle\MongoDBBundle\ManagerRegistry;
 use Doctrine\Common\DataFixtures\Executor\MongoDBExecutor;
 use Doctrine\Common\DataFixtures\Loader;
@@ -11,6 +12,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\ConfirmationQuestion;
+use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\HttpKernel\KernelInterface;
 
 /**
@@ -21,13 +23,22 @@ use Symfony\Component\HttpKernel\KernelInterface;
  */
 class LoadDataFixturesDoctrineODMCommand extends DoctrineODMCommand
 {
+    /**
+     * @var KernelInterface
+     */
     private $kernel;
 
-    public function __construct(ManagerRegistry $registry = null, KernelInterface $kernel = null)
+    /**
+     * @var SymfonyFixturesLoaderInterface
+     */
+    private $fixturesLoader;
+
+    public function __construct(ManagerRegistry $registry = null, KernelInterface $kernel = null, SymfonyFixturesLoaderInterface $fixturesLoader = null)
     {
         parent::__construct($registry);
 
         $this->kernel = $kernel;
+        $this->fixturesLoader = $fixturesLoader;
     }
 
     /**
@@ -45,20 +56,31 @@ class LoadDataFixturesDoctrineODMCommand extends DoctrineODMCommand
             ->setDescription('Load data fixtures to your database.')
             ->addOption('fixtures', null, InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY, 'The directory or file to load data fixtures from.')
             ->addOption('bundles', 'b', InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY, 'The bundles to load data fixtures from.')
+            ->addOption('services', null, InputOption::VALUE_NONE, 'Use services as fixtures')
+            ->addOption('group', null, InputOption::VALUE_IS_ARRAY|InputOption::VALUE_REQUIRED, 'Only load fixtures that belong to this group (use with --services)')
             ->addOption('append', null, InputOption::VALUE_NONE, 'Append the data fixtures instead of flushing the database first.')
             ->addOption('dm', null, InputOption::VALUE_REQUIRED, 'The document manager to use for this command.')
             ->setHelp(<<<EOT
-The <info>doctrine:mongodb:fixtures:load</info> command loads data fixtures from your bundles:
+The <info>doctrine:mongodb:fixtures:load</info> command loads data fixtures from your application:
 
-  <info>./app/console doctrine:mongodb:fixtures:load</info>
+  <info>php %command.full_name%</info>
 
 You can also optionally specify the path to fixtures with the <info>--fixtures</info> option:
-
-  <info>./app/console doctrine:mongodb:fixtures:load --fixtures=/path/to/fixtures1 --fixtures=/path/to/fixtures2</info>
+  <info>php %command.full_name%</info> --fixtures=/path/to/fixtures1 --fixtures=/path/to/fixtures2</info>
 
 If you want to append the fixtures instead of flushing the database first you can use the <info>--append</info> option:
 
-  <info>./app/console doctrine:mongodb:fixtures:load --append</info>
+  <info>php %command.full_name%</info> --append</info>
+
+
+Alternatively, you can also load fixture services instead of files. Fixture services are tagged with `<comment>doctrine.fixture.odm</comment>`.
+When using `<comment>--services</comment>`, both `<comment>--fixtures</comment>` and `<comment>--bundles</comment>` will no longer work.
+Using `<comment>--services</comment>` will be the default behaviour in 4.0.
+When loading fixture services, you can also choose to load only fixtures that live in a certain group:
+`<info>php %command.full_name%</info> <comment>--group=group1</comment> <comment>--services</comment>`
+
+
+
 EOT
         );
     }
@@ -66,11 +88,13 @@ EOT
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $dm = $this->getManagerRegistry()->getManager($input->getOption('dm'));
+        $ui = new SymfonyStyle($input, $output);
 
         $dirOrFile = $input->getOption('fixtures');
         $bundles = $input->getOption('bundles');
-        if ($bundles && $dirOrFile) {
-            throw new \InvalidArgumentException('Use only one option: --bundles or --fixtures.');
+        $services = (bool) $input->getOption('services');
+        if (($services && $bundles) || ($services && $dirOrFile) || ($bundles && $dirOrFile)) {
+            throw new \InvalidArgumentException('Can only use one of "--bundles", "--fixtures", "--services".');
         }
 
         if ($input->isInteractive() && !$input->getOption('append')) {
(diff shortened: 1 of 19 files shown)


DependencyInjection/Compiler/FixturesCompilerPass.php
```diff
@@ -0,0 +1,37 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Doctrine\Bundle\MongoDBBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
```
alcaeus: Useless docblock

DependencyInjection/Compiler/FixturesCompilerPass.php
```diff
@@ -0,0 +1,37 @@
...
+    public function process(ContainerBuilder $container)
+    {
+        $definition = $container->getDefinition('doctrine.mongodb.odm.fixtures.loader');
+        $taggedServices = $container->findTaggedServiceIds(self::FIXTURE_TAG);
+
+        foreach ($taggedServices as $serviceId => $tags) {
+            $definition->addMethodCall('addFixture', [new Reference($serviceId)]);
+        }
+    }
+}
```
alcaeus: There's a missing newline here.

alcaeus:
Sorry for the close/reopen spam - travis-ci needed another nudge to check this PR.

> Do we want to keep depending on the fixturesBundle or should we copy any files over so we can avoid the fixturesBundle which also installs the orm stuff which if you only use odm you don't need? (for example the FixtureGroupInterface)

We don't depend on the fixtures bundle, only on the data-fixtures library, which doesn't have a hard requirement on ORM. Since we don't want to introduce one, copying the necessary files from the fixtures bundle is the way to go for now. I'm afraid I won't be able to implement #509 since it most likely requires changes to the fixtures bundle as well.

> i based it on 3.5.3 as that is what i have locally, it would be nice if this could be included in 3.6, but not sure we can keep it backward compatible? Maybe if we keep fixtures/bundle options and keep their old implementation and if you provide an --autowire option it uses the new one?

If there's going to be a 3.6, it will be in there. 3.6 will happen if I find things dropped in 4.0 that I haven't deprecated in 3.5 yet (I remember there were some). I'll give you an update once I've checked that.

I'll do a comprehensive code review shortly, just wanted to give you a quick update since you opened a pull request so quickly! 🎉

alcaeus:
And a quick update right behind it: there will definitely be a 3.6 release which we'll get going shortly. Once that's done you can change the base on the PR to 3.6 and it will be there 👍

Command/LoadDataFixturesDoctrineODMCommand.php
```diff
@@ -82,37 +84,18 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-            foreach ($this->getKernel()->getBundles() as $bundle) {
-                $paths[] = $bundle->getPath().'/DataFixtures/MongoDB';
-            }
-        }
+        $groups   = $input->getOption('group');
+        $fixtures = $this->fixturesLoader->getFixtures($groups);
+        if (! $fixtures) {
+            $message = 'Could not find any fixture services to load';
 
-        $loaderClass = $this->container->getParameter('doctrine_mongodb.odm.fixture_loader');
```
alcaeus: Dropping this is a BC break: we can change the default value in 3.x but not drop it entirely.

arendjantetteroo: @alcaeus 
Should i add an --autowire option to the command, which if provided it uses the new symfonyLoader?

And if not provided, it uses the old way with the --bundles or --fixtures option? 
And a check that if you pass --autowire, you can't pass --bundles or --fxitures?
And add an @deprecated if you use the old 2? 

Then in 4.0 you can remove the --bundles,--fixtures and old loader?

arendjantetteroo:
> > Do we want to keep depending on the fixturesBundle or should we copy any files over so we can avoid the fixturesBundle which also installs the orm stuff which if you only use odm you don't need? (for example the FixtureGroupInterface)
> 
> We don't depend on the fixtures bundle, only on the data-fixtures library, which doesn't have a hard requirement on ORM. Since we don't want to introduce one, copying the necessary files from the fixtures bundle is the way to go for now. I'm afraid I won't be able to implement #509 since it most likely requires changes to the fixtures bundle as well.

I copied the Fixture class + the fixtureGroupinterface so it works without the FixturesBundle now. 
So it only depends on the doctrine/data-fixtures library.

Command/LoadDataFixturesDoctrineODMCommand.php
alcaeus: The more I think about it, the more I think we need to move this into 4.0. Then you can break BC as much as you want...

arendjantetteroo: The problem is that i can't easily move to 4.0 as lots of code breaks due to the odm2.0 chnges. So if there is a way to get this onto 3.x i can already clean up my fixtures. 
I'll see if i can find an easy way to have both approaches working

arendjantetteroo: Or maybe add a new fixture loader command and deprecate the old one, where in 4.0 you can only use the new one?

arendjantetteroo:
I fixed the issue with the tests by adding a mocked loader, the functionality of the loader is not tested. 
I saw that the FixturesBundle has an integration test for it, should we copy that as well?

arendjantetteroo:
While fixing the command tests i figured out the current loader code is php 7.1 and up with type checking. 

2 options: 
1. Remove all type stuff so 5.6 is still valid (assuming 3.6 should still be 5.x compatible?)
2. Postpone to 4.0

alcaeus:
I'm very inclined to postpone to 4.0 - the trouble doesn't seem to be worth it.

arendjantetteroo:
@alcaeus 
After a few more changes, i've reverted the old way in the command with --bundles and --fixtures. 

If a user wants to use the autowired way, he can use --services. 
The old and new options are incompatible, so if both --bundles/--fixtures and --services are used, an error is thrown indicating it. 
if --services is provided, we assume the user knows that he needs to use the fixtures as services. 
I guess this should be added to the 3.6 upgrade file?
In 4.0 you could then remove the --bundles and --fixtures options if you want to and make this only based on services and the --group feature to define which to load. 

It's working on both my old application way of doing fixtures and working with the new autowired fixtures in a new branch. 

Let me know what else is needed?

Command/LoadDataFixturesDoctrineODMCommand.php
```diff
@@ -45,32 +51,41 @@ protected function configure()
...
     {
         $dm = $this->getManagerRegistry()->getManager($input->getOption('dm'));
+        $ui = new SymfonyStyle($input, $output);
 
         $dirOrFile = $input->getOption('fixtures');
         $bundles = $input->getOption('bundles');
-        if ($bundles && $dirOrFile) {
-            throw new \InvalidArgumentException('Use only one option: --bundles or --fixtures.');
+        $services = $input->getOption('services');
+        if ($bundles && $dirOrFile && $services) {
```
arendjantetteroo: This check should be changed so it checks if more than 1 of the three is used.

alcaeus:
@arendjantetteroo I'll take a deeper look at the PR tonight. If we can keep this entirely BC I'm all for it.

Command/LoadDataFixturesDoctrineODMCommand.php
```diff
@@ -45,32 +56,50 @@ protected function configure()
...
             ->addOption('dm', null, InputOption::VALUE_REQUIRED, 'The document manager to use for this command.')
             ->setHelp(<<<EOT
-The <info>doctrine:mongodb:fixtures:load</info> command loads data fixtures from your bundles:
+The <info>doctrine:mongodb:fixtures:load</info> command loads data fixtures from your application:
 
-  <info>./app/console doctrine:mongodb:fixtures:load</info>
+  <info>php %command.full_name%</info>
 
 You can also optionally specify the path to fixtures with the <info>--fixtures</info> option:
+  <info>php %command.full_name%</info> --fixtures=/path/to/fixtures1 --fixtures=/path/to/fixtures2</info>
```
alcaeus: I would keep original documentation intact except for adding a deprecation notice and add a paragraph below to explain new features. Suggested wording:

> Alternatively, you can also load fixture services instead of files. Fixture services are tagged with `<comment>doctrine.fixture.odm</comment>`.
When using `<comment>--services</comment>`, both `<comment>--fixtures</comment>` and `<comment>--bundles</comment>` will no longer work. Using `<comment>--services</comment>` will be the default behaviour in 4.0.
When loading fixture services, you can also choose to load only fixtures that live in a certain group:
`<info>php %command.full_name%</info> <comment>--group=group1</comment> <comment>--services</comment>`

Command/LoadDataFixturesDoctrineODMCommand.php
```diff
@@ -45,32 +56,50 @@ protected function configure()
...
     {
         $dm = $this->getManagerRegistry()->getManager($input->getOption('dm'));
+        $ui = new SymfonyStyle($input, $output);
 
         $dirOrFile = $input->getOption('fixtures');
         $bundles = $input->getOption('bundles');
-        if ($bundles && $dirOrFile) {
-            throw new \InvalidArgumentException('Use only one option: --bundles or --fixtures.');
+        $services = $input->getOption('services');
+        if ($services === false) {
```
alcaeus: I'd simplify the entire block:
```
if (($services && $bundles) || ($services && $dirOrFile) || ($bundles && $dirOrFile)) {
    throw new \InvalidArgumentException('Can only use one of "--bundles", "--fixtures", "--services".');
}
```

I don't think treating `--services` in combination with `--bundles` or `--fixtures` warrants a warning, I'd just throw a straight up error messages: chances are that people who decide to use `--services` already know what it does.

Command/LoadDataFixturesDoctrineODMCommand.php
```diff
@@ -82,37 +111,51 @@ protected function execute(InputInterface $input, OutputInterface $output)
             }
         }
 
-        if ($dirOrFile) {
-            $paths = is_array($dirOrFile) ? $dirOrFile : [$dirOrFile];
-        } elseif ($bundles) {
-            $paths = [$this->getKernel()->getRootDir().'/DataFixtures/MongoDB'];
-            foreach ($bundles as $bundle) {
-                $paths[] = $this->getKernel()->getBundle($bundle)->getPath();
+        if ($services === false) {
```
alcaeus:
```suggestion
        if (!$services) {
```

Command/LoadDataFixturesDoctrineODMCommand.php
```diff
@@ -45,32 +56,50 @@ protected function configure()
...
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $dm = $this->getManagerRegistry()->getManager($input->getOption('dm'));
+        $ui = new SymfonyStyle($input, $output);
 
         $dirOrFile = $input->getOption('fixtures');
         $bundles = $input->getOption('bundles');
-        if ($bundles && $dirOrFile) {
-            throw new \InvalidArgumentException('Use only one option: --bundles or --fixtures.');
+        $services = $input->getOption('services');
```
alcaeus:
```suggestion
        $services = (bool) $input->getOption('services');
```

Command/LoadDataFixturesDoctrineODMCommand.php
alcaeus: This code path also needs to trigger a deprecation notice explaining that using `--bundles` and `--fixtures` is deprecated and will be removed in 4.0.

Fixture/FixtureGroupInterface.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Doctrine\Bundle\MongoDBBundle\Fixture;
+
+/**
+ * FixtureGroupInterface can be implemented by fixtures that belong in groups
+ */
+interface FixtureGroupInterface
```
alcaeus: I'm not excited to introduce a separate interface for this here, as it means that people that want to use their fixtures for ORM and ODM will have to use two different marker interfaces that provide the same functionality. However, since the corresponding ORM interface resides in the bundle and the library itself doesn't know anything about groups, there might not be a clean way around this 🤔 

I'll have to give this some more thought before merging this.

Resources/config/mongodb.xml
```diff
@@ -210,5 +211,9 @@
 
         <!-- listeners -->
         <service id="doctrine_mongodb.odm.listeners.resolve_target_document" class="%doctrine_mongodb.odm.listeners.resolve_target_document.class%" public="false" />
+
+        <service id="doctrine.mongodb.odm.symfony.fixtures.loader" class="Doctrine\Bundle\MongoDBBundle\Loader\SymfonyFixturesLoader" public="false">
```
alcaeus: Other services are named `doctrine_mongodb.odm...`. Please follow that convention.

Tests/Command/SymfonyFixturesLoaderMock.php
```diff
@@ -0,0 +1,32 @@
...
+    public function addFixtures(array $fixtures)
+    {
+        $this->fixtures = $fixtures;
+    }
+
+    public function getFixtures(array $groups = [])
+    {
+        return $this->fixtures;
+    }
+}
```
alcaeus: Missing newline here.

Tests/Command/SymfonyFixturesLoaderMock.php
```diff
@@ -0,0 +1,32 @@
+<?php
+/**
```
alcaeus: Please drop this comment, it serves no purpose.

Fixture/FixtureGroupInterface.php
arendjantetteroo: The symfony  loaders are also seperate right now because you need the ODM/ORM compiler pass interface, so if you would want to combine the 2 we need to either depend on the fixtureBundle and extend their loader or abstract the loading from the type of database adatper you need and use injection for the ObjectManager in the load() function.

If the fixtures have proper dependency injection, injecting the objectManager in the constructor avoids needing to have the ODM/ORM distinction between the compiler pass marker and you can use just one loader with one interface. 

Abstracting this into a separate bundle makes more sense than including it in the mongodb bundle i guess. 

 I'm not sure how many people use both ORM and OMD anyway? In our project we only use the ODM, so we don't need the FixturesBundle and the assorted doctrine and doctrine cache bundles it takes with it. 

Maybe for 4.0 we could put the fixture stuff in a seperate bundle that depends or extends on the FixturesBundle and override the symfony loader from the fixturesbundle to use the same Fixture interfaces and use dependency injection for the ObjectManager so it can work with both the ORM and ODM stuff. In a newer FixtureBundle they could also move to the injected ObjectManager so you don't need it anymore in the load(). 
It would mean that people that migrate to 3.6 need to change things again for 4.0, but only if they want to combine ORM/ODM. 
It's all a bit messy...

arendjantetteroo:
@alcaeus thanks for the review, updated and rebased. 

Thinking about the ODM/ORM stuff, the only clean way would be extracting the loading of fixtures into a seperate symfony bundle that can work with the same Fixture/FixtureGroup interface and then both the FixtureBundle and this bundle/a new ODMFixture bundle depending on this FixtureLoaderBundle. 

I just don't know if that is worth all the hassle though...

arendjantetteroo:
@alcaeus what's the plan, anything i can do to help?

alcaeus:
> @alcaeus what's the plan, anything i can do to help?

No, I just need to finish looking through this and merge it. I'll try to get that done soon, even if it means copy/pasting large parts of the ORM fixtures bundle. We should clean that up afterwards.

alcaeus:
@arendjantetteroo I've fixed a couple of minor review issues. From a code perspective this looks good, but I think we need to add some tests for this. In DoctrineFixturesBundle, there's an [IntegrationTest](https://github.com/doctrine/DoctrineFixturesBundle/blob/master/Tests/IntegrationTest.php) that we could base our tests on. Do you want to give this a shot or would you rather have me do it? I won't have time for it before the weekend though...

arendjantetteroo:
I'll see what i can do, not sure i get to it before the weekend either :)

arendjantetteroo:
@alcaeus I copy/pasted and then changed the IntegrationTest, works for me locally, waiting for the travis checks now. 
I added a seperate commit for the type removal so it supports 5.6 in 3.6 version of the bundle. 
This commit can then be reverted for 4 to have types if wanted. If we squash it, this info is lost.

arendjantetteroo:
Fails on pre 7.1 with : 
`1) Doctrine\Bundle\MongoDBBundle\Tests\IntegrationTest::testFixturesLoader
Symfony\Component\DependencyInjection\Exception\InvalidArgumentException: Unable to replace alias "doctrine.odm.mongodb.document_manager" with actual definition "doctrine_mongodb.odm.document_manager".`

Any ideas?

arendjantetteroo:
@alcaeus From what i understand i need to load a config into the Extension to get this document manager definition in the service container. However i don't understand why this works fine in 7.1 and 7.2?

alcaeus:
I'm taking a look what's going on - no idea why it isn't working. We'll see 👍

alcaeus:
@arendjantetteroo I inadvertently force-pushed to your remote, overwriting your commits which I wanted to keep. Can you force-push your previous branch over my changes? Thanks!

arendjantetteroo:
@alcaeus I got a message to force push but i think you already corrected it right?

alcaeus:
> @alcaeus I got a message to force push but i think you already corrected it right?

Yeah, I fixed it. Currently trying to fix the alias issue you mentioned above.

PR #531, opened 2019-02-12
Fix registration of the proxy autoloader

### LINKED ISSUE #519: Proxy autoloader is not registered
See https://github.com/doctrine/mongodb-odm/issues/1920 for original discussion. The bundle should automatically register the autoloader for proxies to ensure proxies are not needlessly regenerated.

alcaeus:
Fixes #519 and https://github.com/doctrine/mongodb-odm/issues/1920.

@Rybbow could you give this a spin and let me know if this fixes your issue?

### DIFF
--- a/DoctrineMongoDBBundle.php
+++ b/DoctrineMongoDBBundle.php
@@ -8,21 +8,16 @@ use Doctrine\Bundle\MongoDBBundle\DependencyInjection\Compiler\CreateHydratorDir
 use Doctrine\Bundle\MongoDBBundle\DependencyInjection\Compiler\CreateProxyDirectoryPass;
 use Doctrine\Bundle\MongoDBBundle\DependencyInjection\Compiler\ServiceRepositoryCompilerPass;
 use Doctrine\Bundle\MongoDBBundle\DependencyInjection\DoctrineMongoDBExtension;
+use Doctrine\ODM\MongoDB\Configuration;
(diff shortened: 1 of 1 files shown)


DoctrineMongoDBBundle.php
```diff
@@ -54,56 +49,21 @@ public function getContainerExtension()
...
-
-                    foreach ($classes as $classMetadata) {
-                        if ($classMetadata->name !== $originalClassName) {
-                            continue;
-                        }
-
-                        $dm->getProxyFactory()->generateProxyClasses([$classMetadata]);
-                    }
-                }
+        $this->registerAutoloader($registry->getManager());
```
malarzm: What if there's more than one manager?

alcaeus: I was thinking the same, but the proxy namespace is shared among document managers, so I would think the autoloader would be the same for all managers. I can take a look to see whether there would be any differences.

malarzm: Ah indeed, I thought that each manager can have its own namespace defined 👍

alcaeus: Took another look: since we use the same values of `proxy_dir` and `auto_generate_proxy_classes` for all document managers, checking only the default manager will work as expected for the current configuration. If people decide to override the configuration for a single document manager, they'll have to ensure autoloading for said manager still works as expected.

PR #540, opened 2019-03-13
Use FQCN for services and documents

### LINKED ISSUE #537: Use FQCNs as service identifiers
Leftover from #535

malarzm:
Fixes #537

### DIFF
--- a/Resources/doc/cookbook/registration_form.rst
+++ b/Resources/doc/cookbook/registration_form.rst
@@ -217,6 +217,8 @@ controller for displaying the registration form:
     // src/Acme/AccountBundle/Controller/AccountController.php
     namespace Acme\AccountBundle\Controller;
 
+    use Doctrine\ODM\MongoDB\DocumentManager;
+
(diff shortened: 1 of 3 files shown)


Resources/doc/cookbook/registration_form.rst
```diff
@@ -254,7 +254,7 @@ the validation and saves the data into MongoDB:
 
     public function createAction(Request $request)
     {
-        $dm = $this->get('doctrine_mongodb')->getManager();
+        $dm = $this->get(DocumentManager::class);
```
malarzm: I wasn't fixing `$this->get` calls just yet, but I can do it now, just need to figure out how to show that we're injecting DocumentManager. Or should we do method injection?

alcaeus: I'm all for using method injection: it's enabled by default and it avoids having to add a constructor for DI purposes to the code samples.

malarzm: Done 👍

alcaeus:
Thanks @malarzm!

PR #550, opened 2019-04-07
Event subscriber

emulienfou:
Replacing tag `doctrine_mongodb.odm.event_listener` by `doctrine_mongodb.odm.event_subscriber`.

Following new Pull Requests from the next related issue: https://github.com/doctrine/mongodb-odm/issues/1993

### DIFF
--- a/DependencyInjection/DoctrineMongoDBExtension.php
+++ b/DependencyInjection/DoctrineMongoDBExtension.php
@@ -5,6 +5,7 @@ namespace Doctrine\Bundle\MongoDBBundle\DependencyInjection;
 
 use Doctrine\Bundle\MongoDBBundle\DependencyInjection\Compiler\ServiceRepositoryCompilerPass;
 use Doctrine\Bundle\MongoDBBundle\Repository\ServiceDocumentRepositoryInterface;
+use Doctrine\Common\EventSubscriber;
 use Symfony\Bridge\Doctrine\DependencyInjection\AbstractDoctrineExtension;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Alias;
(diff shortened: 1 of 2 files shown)


DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -74,7 +74,7 @@ public function load(array $configs, ContainerBuilder $container)
             foreach ($config['resolve_target_documents'] as $name => $implementation) {
                 $def->addMethodCall('addResolveTargetDocument', [$name, $implementation, []]);
             }
-            $def->addTag('doctrine_mongodb.odm.event_listener', ['event' => 'loadClassMetadata']);
+            $def->addTag('doctrine_mongodb.odm.event_subscriber');
```
alcaeus: This needs to be updated: we either require ODM 1.3 for version 3.6 of the bundle (which I'd like to avoid), or we check if the service is actually an event subscriber and add the appropriate tags.

DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -74,7 +74,14 @@ public function load(array $configs, ContainerBuilder $container)
             foreach ($config['resolve_target_documents'] as $name => $implementation) {
                 $def->addMethodCall('addResolveTargetDocument', [$name, $implementation, []]);
             }
-            $def->addTag('doctrine_mongodb.odm.event_subscriber');
+
+            // Register service has an event subscriber if implement interface
```
emulienfou: @alcaeus I can only end up doing this to check if service class implements interface. Do you think of a better checking code?

stof: you should not trim the `%`. If any parameter is involved, it should be resolved, not replaced by its name.

Tests/DependencyInjection/AbstractMongoDBExtensionTest.php
```diff
@@ -409,7 +409,14 @@ public function testResolveTargetDocument()
 
         $definition = $container->getDefinition('doctrine_mongodb.odm.listeners.resolve_target_document');
         $this->assertDefinitionMethodCallOnce($definition, 'addResolveTargetDocument', [UserInterface::class, 'MyUserClass', []]);
-        $this->assertEquals(['doctrine_mongodb.odm.event_listener' => [['event' => 'loadClassMetadata']]], $definition->getTags());
+
+        if (in_array('Doctrine\Common\EventSubscriber',
+            class_implements($container->getParameter(trim($definition->getClass(), '%'))))) {
```
stof: Why would the test check that ? Assertions should not be conditional.

emulienfou: Hi @stof actually when executing tests, the branch 1.1 of `doctrine/mongodb-odm` is used.
This branch do not actually have my modifications merged yet.
But I think you are right, I revert the changes and will wait until my 2 PRs are merged.
After the merge, the tests should be green

malarzm:
Assigning to a milestone to keep the stalebot away

Tests/DependencyInjection/AbstractMongoDBExtensionTest.php
alcaeus: I re-added the check. Since we are testing against lowest dependencies, we need to account for two separate versions where either one or the other tag will be added.

alcaeus:
Thanks @emulienfou!

Issue #552, opened 2019-05-31
Symfony 4.2 deprecation : Symfony\Component\HttpKernel\Kernel::getRootDir()


kpn13:
Hi, 

I suppose you already know but still`Symfony\Component\HttpKernel\Kernel::getRootDir()` method is deprecated since sf 4.2

https://github.com/doctrine/DoctrineMongoDBBundle/blob/a896fcf2f50ca8fa253f8d4608192bf50190d8bc/Command/LoadDataFixturesDoctrineODMCommand.php#L94

https://github.com/doctrine/DoctrineMongoDBBundle/blob/a896fcf2f50ca8fa253f8d4608192bf50190d8bc/Command/LoadDataFixturesDoctrineODMCommand.php#L101

malarzm:
@kpn13 thanks for the tip! Would you be willing to work on a PR to get rid of said code?

kpn13:
Ok I can try :)

PR #553, opened 2019-05-31
Replace getRootDir by getProjectDir

kpn13:
See #552


alcaeus:
Thanks for the PR! Just a heads up, we need to fix this in the 3.6 branch, from which I'll merge it up to master. Please also be aware that the logic at hand is changing in #529, so I'll only merge this once that PR is done. I'll rebase this PR when this is the case and make sure there are no conflicts.

PR #554, opened 2019-06-05
Drop file fixtures in favour of fixture services

alcaeus:
After merging #529, we can drop the legacy way of loading fixtures, which is by loading them from a directory or from bundles. Instead, we now only allow loading fixture services.

### DIFF
--- a/Command/LoadDataFixturesDoctrineODMCommand.php
+++ b/Command/LoadDataFixturesDoctrineODMCommand.php
@@ -19,9 +19,6 @@ use Symfony\Component\HttpKernel\KernelInterface;
 use const E_USER_DEPRECATED;
 use function class_exists;
 use function implode;
-use function is_array;
-use function is_dir;
-use function is_file;
 use function sprintf;
 use function trigger_error;
 
@@ -30,9 +27,6 @@ use function trigger_error;
  */
 class LoadDataFixturesDoctrineODMCommand extends DoctrineODMCommand
 {
-    /** @var KernelInterface|null */
-    private $kernel;
-
     /** @var SymfonyFixturesLoaderInterface */
     private $fixturesLoader;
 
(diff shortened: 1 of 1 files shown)


Command/LoadDataFixturesDoctrineODMCommand.php
```diff
@@ -93,11 +84,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $dm = $this->getManagerRegistry()->getManager($input->getOption('dm'));
         $ui = new SymfonyStyle($input, $output);
 
-        $dirOrFile = $input->getOption('fixtures');
-        $bundles   = $input->getOption('bundles');
-        $services  = (bool) $input->getOption('services');
-        if (($services && $bundles) || ($services && $dirOrFile) || ($bundles && $dirOrFile)) {
-            throw new InvalidArgumentException('Can only use one of "--bundles", "--fixtures", "--services".');
+        if ((bool) $input->getOption('services')) {
+            @trigger_error(sprintf('The "services" option to the "%s" command is deprecated and will be dropped in DoctrineMongoDBBundle 5.0.', $this->getName()), E_USER_DEPRECATED);
```
malarzm: I've never used fixtures with ODM (and must admit I haven't watched latest fixtures PRs closely) - why can't we remove this already?

alcaeus:
With 3.6, users will have to use the services option to not have a deprecation message. I want them to only have to worry about removing that once they run 4.0, but still have it work with it instead of adding another BC break.

Command/LoadDataFixturesDoctrineODMCommand.php
alcaeus: With 3.6, users will have to use the services option to not have a deprecation message. I want them to only have to worry about removing that once they run 4.0, but still have it work with it instead of adding another BC break.

PR #556, opened 2019-06-05
Remove mocks to final class in tests

alcaeus:


malarzm:
Thanks, I totally haven't considered there might be mocks in the bundle :)

PR #569, opened 2019-07-12
Fix registration of command subscribers

### LINKED ISSUE #563: Symfony 4 profiler empty
I have no queries on Symfony profiler when changing to 4.0.0-RC1

alcaeus:
Fixes #563. Command subscribers were not properly registered - just adding a method call to the service definition is not enough.

To work around this, we add a command subscriber registry that is registered and unregistered when the bundle is booted and shut down. The registry proxies all calls to `register` and `unregister` to all registered subscribers. Since we don't have an interface for registrable command subscribers, I simply added `method_exists` checks to avoid fatal errors. We can still add an interface for such subscribers to the next release candidate for mongodb-odm and change the code to throw exceptions instead of silently skipping incompatible subscribers. Opinions @jmikola @malarzm?

### DIFF
--- a/APM/CommandLoggerRegistry.php
+++ b/APM/CommandLoggerRegistry.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Bundle\MongoDBBundle\APM;
+
+use Doctrine\ODM\MongoDB\APM\CommandLoggerInterface;
+use function array_map;
+
+final class CommandLoggerRegistry
+{
+    /** @var CommandLoggerInterface[] */
+    private $commandLoggers = [];
+
+    public function __construct(iterable $commandLoggers)
+    {
+        foreach ($commandLoggers as $commandLogger) {
+            $this->addLogger($commandLogger);
+        }
+    }
+
+    public function register() : void
+    {
+        array_map(static function (CommandLoggerInterface $commandLogger) {
+            $commandLogger->register();
+        }, $this->commandLoggers);
+    }
+
+    public function unregister() : void
+    {
+        array_map(static function (CommandLoggerInterface $commandLogger) {
+            $commandLogger->unregister();
+        }, $this->commandLoggers);
+    }
+
+    private function addLogger(CommandLoggerInterface $logger) : void
+    {
+        $this->commandLoggers[] = $logger;
+    }
+}
--- a/DoctrineMongoDBBundle.php
+++ b/DoctrineMongoDBBundle.php
@@ -55,6 +55,7 @@ class DoctrineMongoDBBundle extends Bundle
         $registry = $this->container->get('doctrine_mongodb');
 
         $this->registerAutoloader($registry->getManager());
+        $this->registerCommandLoggers();
     }
 
     private function registerAutoloader(DocumentManager $documentManager) : void
@@ -69,7 +70,7 @@ class DoctrineMongoDBBundle extends Bundle
         spl_autoload_register($this->autoloader);
     }
 
-    public function shutdown()
+    private function unregisterAutoloader() : void
     {
         if ($this->autoloader === null) {
             return;
@@ -77,5 +78,23 @@ class DoctrineMongoDBBundle extends Bundle
 
         spl_autoload_unregister($this->autoloader);
         $this->autoloader = null;
+    }
+
+    private function registerCommandLoggers() : void
+    {
+        $commandLoggerRegistry = $this->container->get('doctrine_mongodb.odm.command_logger_registry');
+        $commandLoggerRegistry->register();
+    }
+
+    private function unregisterCommandLoggers() : void
+    {
+        $commandLoggerRegistry = $this->container->get('doctrine_mongodb.odm.command_logger_registry');
+        $commandLoggerRegistry->unregister();
+    }
+
+    public function shutdown()
+    {
+        $this->unregisterAutoloader();
+        $this->unregisterCommandLoggers();
     }
 }
--- a/Tests/FixtureIntegrationTest.php
+++ b/Tests/FixtureIntegrationTest.php
@@ -41,7 +41,7 @@ class FixtureIntegrationTest extends TestCase
 
     public function testFixturesLoader() : void
     {
-        $kernel = new IntegrationTestKernel('dev', true);
+        $kernel = new IntegrationTestKernel('dev', false);
         $kernel->addServices(static function (ContainerBuilder $c) : void {
             $c->autowire(OtherFixtures::class)
                 ->addTag(FixturesCompilerPass::FIXTURE_TAG);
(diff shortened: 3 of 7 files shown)


APM/CommandSubscriberRegistry.php
```diff
@@ -0,0 +1,41 @@
...
+use function method_exists;
+
+final class CommandSubscriberRegistry
+{
+    /** @var CommandSubscriber[] */
+    private $commandSubscribers;
+
+    public function __construct(iterable $commandSubscribers)
+    {
+        $this->commandSubscribers = $commandSubscribers;
```
jmikola: Would it make sense for this to validate that each element within `$commandSubscribers` is actually a CommandSubscriber instance?

APM/CommandSubscriberRegistry.php
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public function register() : void
+    {
+        foreach ($this->commandSubscribers as $subscriber) {
+            if (! method_exists($subscriber, 'register')) {
+                continue;
+            }
+
+            $subscriber->register();
```
jmikola: Since [CommandSubscriber](https://www.php.net/manual/en/class.mongodb-driver-monitoring-commandsubscriber.php) has no `register()` and `unregister()` methods, I don't believe we'd ever reach this point. Subscribers are registered with the extension globally via [these functions](https://www.php.net/manual/en/ref.monitoring.functions.php).

The [APM tutorial](https://www.php.net/manual/en/mongodb.tutorial.apm.php) has some examples that tie this all together.

APM/CommandSubscriberRegistry.php
jmikola: On second thought, I wonder if this is instead supposed to accept only [`Doctrine\ODM\MongoDB\APM\CommandLogger`](https://github.com/doctrine/mongodb-odm/blob/6ded0705dba3661438b501bfaa39f42858293137/lib/Doctrine/ODM/MongoDB/APM/CommandLogger.php) instances, as that class is where `register()` and `unregister()` methods are defined.

jmikola:
> Since we don't have an interface for registrable command subscribers, I simply added method_exists checks to avoid fatal errors. We can still add an interface for such subscribers to the next release candidate for mongodb-odm and change the code to throw exceptions instead of silently skipping incompatible subscribers.

This might be related to my last comment. When I initially reviewed this I was not aware of the [CommandLogger](https://github.com/doctrine/mongodb-odm/blob/6ded0705dba3661438b501bfaa39f42858293137/lib/Doctrine/ODM/MongoDB/APM/CommandLogger.php) class in ODM.

Was the [CommandSubscriber](https://www.php.net/manual/en/class.mongodb-driver-monitoring-commandsubscriber.php) type hint purposeful?

I understand if you don't want to add an interface with `register()`/`unregister()` methods for the ODM subscriber in this PR, but I think it may be worth throwing an exception out of the gate so listeners are never silently ignored. Alternatively, you could manually check for CommandLogger instances when constructing the registry; however, that may be a no-go if the intention is to avoid checking classes at all costs (favoring interfaces).

alcaeus:
Looking at the comments, it definitely makes more sense to introduce our interface for loggers, which only declares the `register` and `unregister` methods.

However, the fact that command subscriber don't only react to commands sent through the current client but rather collect all commands executed on the server makes me doubt whether we want to add the logging functionality in this form. I believe it is definitely unexpected and potentially dangerous to log all commands executed on the server in the application. Since the events also don't expose any information on who triggered the command, we most likely won't be able to filter out any unwanted commands either.

markitosgv:
Any updates? Thanks!!

alcaeus:
> Any updates? Thanks!!

Not yet.

APM/CommandSubscriberRegistry.php
alcaeus: We'll have to introduce a new interface, as the `CommandLogger` class is `final` but we also need a command logger that logs to any PSR-compatible logger. I'd introduce a `CommandLogger` interface that extends `CommandSubscriber` and exposes `register` and `unregister` methods, then limit possible arguments to this method to that interface.

alcaeus:
Updated the code with the new `CommandLoggerInterface` contained in RC2 of ODM. This can be merged once RC2 is released, at which point I'll also remove the commit requiring dev-master.

APM/CommandLoggerRegistry.php
```diff
@@ -0,0 +1,40 @@
...
+    public function register() : void
+    {
+        array_map(static function (CommandLoggerInterface $commandLogger) {
+            $commandLogger->register();
+        }, $this->commandLoggers);
+    }
+
+    public function unregister() : void
+    {
+        array_map(static function (CommandLoggerInterface $commandLogger) {
```
malarzm: Foreach is bad? :D

jmikola: One benefit is that this applies type checking to each element, which would otherwise require an `instanceof` check within `foreach`.

DoctrineMongoDBBundle.php
```diff
@@ -78,4 +79,22 @@ public function shutdown()
...
+        $commandLoggerRegistry->register();
+    }
+
+    private function unregisterCommandLoggers() : void
+    {
+        $commandLoggerRegistry = $this->container->get('doctrine_mongodb.odm.command_logger_registry');
+        $commandLoggerRegistry->unregister();
+    }
+
+    public function shutdown()
```
jmikola: Should this have a return type hint, or is that omitted because we're overriding a method from Symfony's base class?

composer.json
```diff
@@ -13,7 +13,7 @@
     "require": {
         "php": "^7.2",
         "doctrine/annotations": "^1.2",
-        "doctrine/mongodb-odm": "^2.0.0-beta1",
+        "doctrine/mongodb-odm": "dev-master",
```
jmikola: Should this be changed to 2.0.0-RC2 (following its release, of course)?

EDIT: Just saw the "REMOVE:" commit. Makes sense.

Tests/FixtureIntegrationTest.php
```diff
@@ -41,7 +41,7 @@ protected function setUp()
 
     public function testFixturesLoader() : void
     {
-        $kernel = new IntegrationTestKernel('dev', true);
+        $kernel = new IntegrationTestKernel('dev', false);
```
jmikola: Is this pertinent to the PR, or just some extra cleanup?

DoctrineMongoDBBundle.php
alcaeus: I've omitted it because Symfony doesn't have one and I haven't had time to test this against Symfony 5 so far. All Doctrine Bundles will get the Symfony 4.4/5.0 treatment after the Symfony Code Freeze next week (DoctrineBundle has been the guinea pig for this). This will add some return type annotations where applicable. I'll have to see how to treat this since I don't want to hold 4.0 back until November, but I also don't want to release a new major version shortly after releasing 4.0 already.

Tests/FixtureIntegrationTest.php
alcaeus: This was some necessary cleanup as tests started failing.

APM/CommandLoggerRegistry.php
alcaeus: That was one of the reasons why I added it. Of course, a type check is already done in the constructor when storing the items, but then I've started using `array_map` in place of `foreach` to make it more explicit. I can change it to a loop if you prefer that

PR #577, opened 2019-08-20
Changed 'exit()' calls to throw RuntimeException instead

ksowa:
I think this is rather important to give programmers/dev-ops some more feedback than just a message returned in `exit()` function. With exceptions we give the app a chance to put the error message into some log file.

### DIFF
--- a/DependencyInjection/Compiler/CreateHydratorDirectoryPass.php
+++ b/DependencyInjection/Compiler/CreateHydratorDirectoryPass.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\Bundle\MongoDBBundle\DependencyInjection\Compiler;
 
+use RuntimeException;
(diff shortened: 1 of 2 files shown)


stof:
this is also much better, as it can go through any error reporting system that the project has, rather than silently exiting the process.

alcaeus:
Oh wow, didn't know we had `exit` calls in there. Could you please rebase the changes on top of the 3.5 branch and change the base of the pull request? I'm flagging this as a bug to be fixed in 3.5.

ksowa:
@alcaeus I assume you meant 3.6 branch?

alcaeus:
3.5 please - 3.6 is an upcoming release that I'll merge up to.

ksowa:
@alcaeus I've created a new PR - https://github.com/doctrine/DoctrineMongoDBBundle/pull/578

PR #582, opened 2019-09-24
driver options overridden because of wrong key used

lljaworski:


alcaeus:
This is just in time for RC2 later today. I've taken your commit and added a few tests, so I'll close this PR in favour of #583. Thanks @lljaworski!

PR #586, opened 2019-09-30
Fix deprecations when running against ODM 1.3

alcaeus:
Apparently, some deprecations slipped through the cracks.

* Indirect notices are skipped - we generally can't fix these (think vendor code triggering deprecation notices)
* A single usage of a deprecated document repository class is ignored as fixing it requires bumping the minimum ODM version (which can't be done in a patch release)
* All other deprecations were fixed in the test suite.

### DIFF
--- a/Tests/Mapping/Driver/YamlDriverTest.php
+++ b/Tests/Mapping/Driver/YamlDriverTest.php
@@ -5,6 +5,9 @@ namespace Doctrine\Bundle\MongoDBBundle\Tests\Mapping\Driver;
 
 use Doctrine\Bundle\MongoDBBundle\Mapping\Driver\YamlDriver;
 
+/**
+ * @group legacy
+ */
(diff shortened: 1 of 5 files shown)


Tests/Mapping/Driver/YamlDriverTest.php
```diff
@@ -5,6 +5,9 @@
 
 use Doctrine\Bundle\MongoDBBundle\Mapping\Driver\YamlDriver;
 
+/**
+ * @group legacy
```
malarzm: this seems unrelated, is it some magic group name that allows deprecations to be present?

alcaeus: The group tells the Symfony deprecation handler to ignore all deprecation notices that come up during the execution of this test. This can be done to specifically test legacy behaviour without having to define an exception for the deprecation in configuration.

PR #592, opened 2019-10-08
Authentication details

AntoniusGolly:
As true for mysql you can set up users in mongo to increase the security. When this is done one needs to provide the ODM driver with the information of username, password and authentication database (the database within the connection where the actual user is stored, which is different from the default database. I found after some intense research how it's done and found out it is not mentioned in the docs so I wanted to add it here were it belongs.

### DIFF
--- a/Resources/doc/installation.rst
+++ b/Resources/doc/installation.rst
@@ -78,6 +78,21 @@ the MongoDB ODM across your application:
     Of course, you'll also need to make sure that the MongoDB server is running
     in the background. For more details, see the MongoDB `Quick Start`_ guide.
 
+
+Authentication
+--------------
+
+If you use authentication on your MongoDB database you can the provide username, 
+password, and authentication database in the following way:
+
+    # app/config/parameters.yaml
+    parameters:
+        mongodb_server: "mongodb://username:password@localhost:27017/?authSource=auth-db"
+
+.. note::
+
+    The authentication database is different to the default database used by MongoDB.
+
 .. _`installation chapter`: https://getcomposer.org/doc/00-intro.md
 .. _`MongoDB driver`: https://docs.mongodb.com/ecosystem/drivers/php/
 .. _`Quick Start`: http://www.mongodb.org/display/DOCS/Quickstart


Resources/doc/installation.rst
```diff
@@ -78,6 +78,19 @@ the MongoDB ODM across your application:
     Of course, you'll also need to make sure that the MongoDB server is running
     in the background. For more details, see the MongoDB `Quick Start`_ guide.
 
+
+###Authorization
```
malarzm:
```suggestion
Configuration
-------------
```

This is a RST file so this is the correct header

Resources/doc/installation.rst
```diff
@@ -78,6 +78,19 @@ the MongoDB ODM across your application:
...
+
+If you use authorization on your mongo database you can provide username, password and authentication database in the following way:
```
malarzm: Please change all occurrences of "mongo" to "MongoDB". Also there should be a comma after "password"

Resources/doc/installation.rst
AntoniusGolly: I guess `Configuration` in your suggestion should be replaced with `Authorization`

Resources/doc/installation.rst
```diff
@@ -78,6 +78,20 @@ the MongoDB ODM across your application:
     Of course, you'll also need to make sure that the MongoDB server is running
     in the background. For more details, see the MongoDB `Quick Start`_ guide.
 
+
+Authorization
```
jmikola: This should be [Authentication](https://docs.mongodb.com/manual/core/authentication/). "Authorization" would more correctly pertain to [role-based access control](https://docs.mongodb.com/manual/core/authorization/), which doesn't really involve the driver since roles are configured entirely on the server. This quote from the manual succinctly explains the difference between the two:

> Authentication is the process of verifying the identity of a client. When access control, i.e. authorization, is enabled, MongoDB requires all clients to authenticate themselves in order to determine their access.

Resources/doc/installation.rst
```diff
@@ -78,6 +78,20 @@ the MongoDB ODM across your application:
...
+-------------
+
+If you use authorization on your MongoDB database you can provide username, password, and authentication database in the following way:
+
+    # app/config/parameters.yml
+    parameters:
+        mongodb_server: "mongodb://username:password@localhost:27017/auth-db"
```
jmikola: The [authSource](https://docs.mongodb.com/manual/reference/connection-string/#urioption.authSource) URI option is the preferred mechanism for specifying the authentication database. I'd suggest replacing `/auth-db` with `/?authSource=auth-db` here.

The `/auth-db` would actually specify both a default database _and_ default authentication database. Most drivers (including PHP) have no concept of a default database, since their API typically requires users to explicitly select a database and collection through the client object. Doctrine ODM does have a default database (when unspecified in model mappings), but I don't recall if it parses it out of the connection string (perhaps @malarzm can confirm). In any event, `authSource` is more explicit and also avoids any ambiguity in the event that Doctrine _does_ parse the URI and a user might need to provide distinct values for both.

I'll also point out that for some auth mechanisms (e.g. LDAP, Kerberos), the value of `authSource` does not represent a database name at all, so that is another reason not to rely on the overloaded `/auth-db` component in modern applications.

See [Connection String: Components](https://docs.mongodb.com/manual/reference/connection-string/#components) in the MongoDB manual for more information.

Resources/doc/installation.rst
malarzm: Right! Sorry, it's too late :)

Resources/doc/installation.rst
```diff
@@ -78,6 +78,20 @@ the MongoDB ODM across your application:
     Of course, you'll also need to make sure that the MongoDB server is running
     in the background. For more details, see the MongoDB `Quick Start`_ guide.
 
+
+Authorization
+-------------
+
+If you use authorization on your MongoDB database you can provide username, password, and authentication database in the following way:
```
jmikola: "authorization" should be "authentication" here as well.

Resources/doc/installation.rst
```diff
@@ -78,6 +78,20 @@ the MongoDB ODM across your application:
...
+
+    # app/config/parameters.yml
+    parameters:
+        mongodb_server: "mongodb://username:password@localhost:27017/auth-db"
+
+.. note::
+   
+    The authentication database is different to the default database and needs to be provided as MongoDB checks during aurhorization the users collection of that particular database.
```
jmikola: I think this note can be removed if the above changes are made. If kept, I'd suggest rephrasing as:

> The authentication database is different from the default database used by MongoDB ODM.

From an ODM user's respective, I think the concept of a default database is when asking the question "where does my application and model data live?". Also, the relevant collection is [system.users](https://docs.mongodb.com/manual/reference/system-users-collection/index.html) in current server versions, but again that only applies to auth mechanisms that actually use a database in lieu of some external service.

Resources/doc/installation.rst
AntoniusGolly: Agreed, will change in the next commit.

Resources/doc/installation.rst
AntoniusGolly: Thanks for the background. I tested the connection with the url param notation and it works.

AntoniusGolly:
bump

malarzm:
@AntoniusGolly could you please squash your commits into one? I'll merge this right after you do so :)

Resources/doc/installation.rst
```diff
@@ -78,6 +78,20 @@ the MongoDB ODM across your application:
     Of course, you'll also need to make sure that the MongoDB server is running
     in the background. For more details, see the MongoDB `Quick Start`_ guide.
 
+
+Authentication
+--------------
+
+If you use authentication on your MongoDB database you can provide username, password, and authentication database in the following way:
```
greg0ire: Please comply with the sf docs standards (I think we use something similar for Doctrine):

> [Each line should break approximately after the first word that crosses the 72nd character (so most lines end up being 72-78 characters);](http://symfony.com/doc/current/contributing/documentation/standards.html)

Resources/doc/installation.rst
```diff
@@ -78,6 +78,20 @@ the MongoDB ODM across your application:
...
+
+    # app/config/parameters.yml
```
greg0ire:
```suggestion
    # app/config/parameters.yaml
```

AntoniusGolly:
@malarzm I squashed into commit "final" but there  appears another commit. Do you want me to fix before merging?

.gitignore
```diff
@@ -3,3 +3,6 @@ composer.phar
 composer.lock
 phpunit.xml
 .phpcs-cache
+.idea
+.history
+.gitignore
```
alcaeus: Please remove changes to this file: `.idea` and `.history` should be handled in your personal ignore file (see https://help.github.com/en/github/using-git/ignoring-files#create-a-global-gitignore) as adding a bunch of IDE configs doesn't make sense for projects. `.gitignore` just shouldn't be there at all ;)

malarzm:
@AntoniusGolly there should be only 1 commit with all changes, right now there's 6. If you need help I can guide you through the process :)

Resources/doc/installation.rst
```diff
@@ -82,9 +82,10 @@ the MongoDB ODM across your application:
 Authentication
 --------------
 
-If you use authentication on your MongoDB database you can provide username, password, and authentication database in the following way:
+If you use authentication on your MongoDB database you can provide username,
```
jmikola: Small grammar suggestion: "you can provide the username, password, ..."

I think that will read better.

Resources/doc/installation.rst
```diff
@@ -78,6 +78,21 @@ the MongoDB ODM across your application:
...
+If you use authentication on your MongoDB database you can provide username, 
+password, and authentication database in the following way:
+
+    # app/config/parameters.yaml
+    parameters:
+        mongodb_server: "mongodb://username:password@localhost:27017/?authSource=auth-db"
+
+.. note::
+
+    The authentication database is different to the default database used by MongoDB.
```
greg0ire:
```suggestion
    The authentication database is different from the default database used by MongoDB.
```

malarzm: From what I've googled both forms are correct

greg0ire: Oh indeed! Sorry.

malarzm:
@AntoniusGolly thank you very much for the addition and patience!

PR #594, opened 2019-10-18
Add support for symfony/stopwatch component

veteran29:
Basic support for profiling via symfony/stopwatch:
image

### DIFF
--- a/APM/PSRCommandLogger.php
+++ b/APM/PSRCommandLogger.php
@@ -28,10 +28,10 @@ final class PSRCommandLogger implements CommandLoggerInterface
     /** @var Stopwatch */
     private $stopwatch;
 
-    public function __construct(?LoggerInterface $logger, string $prefix = 'MongoDB command: ', Stopwatch $stopwatch = null)
+    public function __construct(?LoggerInterface $logger, string $prefix = 'MongoDB command: ', ?Stopwatch $stopwatch = null)
     {
-        $this->logger = $logger;
-        $this->prefix = $prefix;
+        $this->logger    = $logger;
+        $this->prefix    = $prefix;
         $this->stopwatch = $stopwatch;
     }
 
@@ -61,7 +61,7 @@ final class PSRCommandLogger implements CommandLoggerInterface
             return;
         }
 
-        if (null !== $this->stopwatch) {
+        if ($this->stopwatch !== null) {
             $this->stopwatch->start('mongodb', 'mongodb');
         }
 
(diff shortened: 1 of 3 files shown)


APM/PSRCommandLogger.php
```diff
@@ -56,14 +61,24 @@ public function commandStarted(CommandStartedEvent $event)
             return;
         }
 
+        if ($this->stopwatch !== null) {
+            $this->stopwatch->start('mongodb', 'mongodb');
```
alcaeus: I would suggest using the command name here to make the data more informative. Same in the corresponding `stop` calls.

veteran29: Maybe with a `mongodb_` prefix?

It looks like this right now:
image

APM/PSRCommandLogger.php
```diff
@@ -56,14 +61,24 @@ public function commandStarted(CommandStartedEvent $event)
             return;
         }
 
+        if ($this->stopwatch !== null) {
+            $this->stopwatch->start($event->getCommandName(), 'mongodb');
```
jmikola: I think this is preferable to the earlier suggestion about prefixing the name with `mongodb_`, as the category argument will ensure events are grouped (even if the default UI only displays text for the command name itself by default).

It's interesting that `stop()` doesn't also take a category name. That makes me wonder how it determines that we're closing the same event and whether we could possibly conflict if some other handler used a generic event name (e.g. "find", "update"); however, I haven't worked with the component directly so perhaps this isn't a concern.

Resources/config/mongodb.xml
```diff
@@ -138,6 +139,8 @@
 
         <service id="doctrine_mongodb.odm.command_logger" class="Doctrine\Bundle\MongoDBBundle\APM\PSRCommandLogger" public="false">
             <argument type="service" id="logger" on-invalid="null" />
+            <argument>%doctrine_mongodb.odm.command_logger.prefix%</argument>
```
jmikola: @alcaeus: If this is now part of the service definition, would it make sense to change the default value in the constructor signature to `null`?

APM/PSRCommandLogger.php
alcaeus: @jmikola the `stop` method always works on the last open section, not on a specific one. 🤷‍♂️

Resources/config/mongodb.xml
alcaeus: I don't like using DI parameters for this, at least not to have this configurable to the outside. Instead, this should be configured using the configuration and then injected into the DI container by the extension.

PR #602, opened 2019-11-20
Fix symfony 5 compatibility

### LINKED ISSUE #596: Symfony 4.4 deprecation / Symfony 5.0 compatibility
> Not declaring the "Symfony\Bridge\Doctrine\DependencyInjection\AbstractDoctrineExtension::getMetadataDriverClass" method in class "Doctrine\Bundle\MongoDBBundle\DependencyInjection\DoctrineMongoDBExtension" is deprecated since Symfony 4.4. This method will be abstract in Symfony 5.0.

Spotted in https://github.com/api-platform/core/pull/3009

chalasr:
Fixes #596

### DIFF
--- a/DependencyInjection/DoctrineMongoDBExtension.php
+++ b/DependencyInjection/DoctrineMongoDBExtension.php
@@ -420,6 +420,11 @@ class DoctrineMongoDBExtension extends AbstractDoctrineExtension
         return 'mongodb';
     }
 
+    protected function getMetadataDriverClass(string $driverType) : string
+    {
+        return '%' . $this->getObjectManagerElementName('metadata.' . $driverType . '.class') . '%';
+    }
+
     public function getAlias()
     {
         return 'doctrine_mongodb';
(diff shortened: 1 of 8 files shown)


alcaeus:
@chalasr thank you for the PR. I'll adopt the build file from DoctrineBundle to test multiple Symfony versions here and merge this once that is complete.

DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -420,6 +420,11 @@ protected function getMappingResourceExtension()
         return 'mongodb';
     }
 
+    protected function getMetadataDriverClass(string $driverType) : string
+    {
+        return '%' . $this->getObjectManagerElementName('metadata.' . $driverType . '.class%');
```
SenseException: Little nitpick: I would prefer the `%` at the end outside of the method `getObjectManagerElementName()`. The call depends on the implementation that the argument will be appended on whatever the method does inside.

```suggestion
        return '%' . $this->getObjectManagerElementName('metadata.' . $driverType . '.class') . '%';
```

alcaeus:
Thanks @chalasr!

PR #611, opened 2020-01-21
Add server selection timeout and try once options

dzoeteman:
The `serverSelectionTryOnce` and `serverSelectionTimeoutMS` options are not available in the configuration, even though they do work for the MongoDB client. 

This PR adds these two options, which are useful when dealing with replicaset failovers.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -301,6 +301,8 @@ class Configuration implements ConfigurationInterface
                                             })->thenInvalid('The replicaSet option must be a string or null.')
                                         ->end()
                                     ->end()
+                                    ->integerNode('serverSelectionTimeoutMS')->end()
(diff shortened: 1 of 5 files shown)


alcaeus:
I'm not sure if we want to keep adding options to this list. There is a plethora of options available that can be passed via the connection string, which is what I'd recommend is done in this case. Is there a reason why this doesn't work for you?

dzoeteman:
@alcaeus Haven't tried with a connection string yet, though we prefer this method of config as it's easier to inject different separate parameters via environment variables (very useful when running in docker). For example, in all our environments our mongodb hostname is the same, but in development we aren't running a replicaset. 

If you do really want to focus options to the connection string, I think it would be good to keep the options to a minimal, as otherwise folks like us try to use the config parameters and get stuck when not all options are available :)

PR #613, opened 2020-01-31
Introduce DOCtor-RST

OskarStark:
DOCtor-RST is a linting tool (which runs as GithubAction), which is used at symfony-docs, Sonata and EasyAdminBundle.

It checks the rst files for some specific conventions.

You can see it in Action here: https://github.com/symfony/symfony-docs/pull/13006/checks?check_run_id=414955239

We are using it since a few months for the symfony-docs and is developed by me 😄 

https://github.com/OskarStark/doctor-rst

### DIFF
--- a/Resources/doc/.doctor-rst.yaml
+++ b/Resources/doc/.doctor-rst.yaml
@@ -21,7 +21,6 @@ rules:
     php_prefix_before_bin_console: ~
     use_deprecated_directive_instead_of_versionadded: ~
     space_before_self_xml_closing_tag: ~
-    no_explicit_use_of_code_block_php: ~
     ensure_order_of_code_blocks_in_configuration_block: ~
     lowercase_as_in_use_statements: ~
     ordered_use_statements: ~
--- a/Resources/doc/.doctor-rst.yaml
+++ b/Resources/doc/.doctor-rst.yaml
@@ -1,0 +1,48 @@
+rules:
+    no_inheritdoc: ~
+    avoid_repetetive_words: ~
+    blank_line_after_directive: ~
+    short_array_syntax: ~
+    typo: ~
+    replacement: ~
+    versionadded_directive_should_have_version: ~
+    deprecated_directive_should_have_version: ~
+    max_blank_lines:
+        max: 2
+    american_english: ~
+    valid_use_statements: ~
+    no_composer_req: ~
+    composer_dev_option_not_at_the_end: ~
+    no_php_open_tag_in_code_block_php_directive: ~
+    no_blank_line_after_filepath_in_php_code_block: ~
+    no_blank_line_after_filepath_in_yaml_code_block: ~
+    no_blank_line_after_filepath_in_xml_code_block: ~
+    no_blank_line_after_filepath_in_twig_code_block: ~
+    php_prefix_before_bin_console: ~
+    use_deprecated_directive_instead_of_versionadded: ~
+    space_before_self_xml_closing_tag: ~
+    no_explicit_use_of_code_block_php: ~
+    ensure_order_of_code_blocks_in_configuration_block: ~
+    lowercase_as_in_use_statements: ~
+    ordered_use_statements: ~
+    no_namespace_after_use_statements: ~
+    replace_code_block_types: ~
+    use_https_xsd_urls: ~
+    blank_line_before_directive: ~
+    extension_xlf_instead_of_xliff: ~
+    valid_inline_highlighted_namespaces: ~
+    indention: ~
+    unused_links: ~
+    no_app_bundle: ~
+
+    versionadded_directive_major_version:
+        major_version: 3
+
+    versionadded_directive_min_version:
+        min_version: '3.0'
+
+    deprecated_directive_major_version:
+        major_version: 3
+
+    deprecated_directive_min_version:
+        min_version: '3.0'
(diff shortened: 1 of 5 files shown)


alcaeus:
@OskarStark could you please add a comment to tell me what DOCtor-RST does? ;)

OskarStark:
Sure, I updated the PR header 👍

Resources/doc/config.rst
```diff
@@ -343,17 +343,13 @@ following syntax:
             </doctrine_mongodb:config>
         </container>
 
-Now you can retrieve the configured services connection services:
-
-.. code-block:: php
```
malarzm: Why is this a bad thing?

OskarStark: It’s not called but more noise then you need and it reads more fluently.
Also you have highlighted code block in PHPStorm for example

malarzm: Ah ok, I thought it was the `php` part that would allow IDEs to understand there's PHP (not brainfuck for one) inside the code block

OskarStark:
Thank you 😊 my plan is to add it to all doctrine repositories 😊✌🏻

SenseException:
The website only reads the docs of repository branches from the .doctrine-project.json file on master. Maybe I can test it locally in this case.

OskarStark:
Hi @SenseException can you tell me which Sphinx version you are using or can I see your setup?

We at Sonata-Project and Symfony (docs) added a GithubActions workflow to build the docs and see if its working. Are you interested in this too? If yes I could provide one in a separate PR 👍

alcaeus:
> We at Sonata-Project and Symfony (docs) added a GithubActions workflow to build the docs and see if its working. Are you interested in this too?

Yes please. 😍

SenseException:
The Doctrine website is currently using https://github.com/doctrine/rst-parser for creating the docs, but having Sphinx would be great too. It can create a HTML documentation pretty fast for a big amount of ReStructuredText files.

OskarStark:
Hmm I have no experience with the rst parser. Does my PR works after dropping this commit? https://github.com/doctrine/DoctrineMongoDBBundle/commit/ec846d5ee00158551c908593ce9c65a0c227e490

Can you do and try or shall I drop the commit from this PR?

Or maybe just tell the parser to render twig the same as jinja? :thinking:

SenseException:
For now please use `html+jinja` again. Did DOCtor marked this as an issue?

I wrote a small example from the parser and pushed it into https://github.com/SenseException/rst-builder. I think it misses a few implementations that got directly implemented into https://github.com/doctrine/doctrine-website itself where the documentation of every version is built. If Sphinx will at one point of time be chosen as a new rst-builder, it would be needed in the build steps of the Doctrine website.

OskarStark:
Ok I will fix it tomorrow.
Yes, it was initially developed for the Symfony docs and we prefer to use
twig instead of jinja though... not an issue but more or less our standard,
but we can disable the rule ofc 👍🏻

Claudio Zizza <notifications@github.com> schrieb am Di. 11. Feb. 2020 um
21:11:

OskarStark:
Done ✅ can you please check again if the rst-parser is complaining about sth? Thanks 🙏

SenseException:
It's looking good so far. The code blocks where you replaced `.. code-block:: php` with `::` will result in missing highlighting and missing line numbers.

This is how it currently looks like:
docs-orig

And this how it looks like with your changes:
docs-new

OskarStark:
Hi @SenseException,

thank you for your feedback. I cannot say much about the RST-Parser of doctrine and we had the same "problem" at the symfony-docs repo, to not have highlighting when using `::`. We fixed it by telling sphinx "Hey we do mostly PHP stuff, so use it by default" in our [config](https://github.com/symfony/symfony-docs/blob/master/_build/conf.py#L130)

We can ofc. keep it like before and deactivate the rule or switch to Sphinx, but I don't know which advantages doctrine/rst-parser has at the moment

malarzm:
@OskarStark could you take a look at https://github.com/doctrine/DoctrineMongoDBBundle/pull/613#issuecomment-585396044 ? Do we need to use a default language setting somewhere so the code block won't lose their formatting?

OskarStark:
I just answered 1 minute before 😄

malarzm:
Hah perfect timing :) IIRC RST-Parser was introduced to get rid of sphinx as we were moving out from readthedocs. If we could keep the language setting per code block and it won't bother Symfony ecosystem that'd be great.

OskarStark:
> If we could keep the language setting per code block and it won't bother Symfony ecosystem that'd be great.

Sure thing, it will not bother us, its just (because we have a lot of docs), less distracting/noisy to read, I will disable the rule and revert the commit

Resources/doc/.doctor-rst.yaml
```diff
@@ -0,0 +1,47 @@
...
+        major_version: 3
+
+    versionadded_directive_min_version:
+        min_version: '3.0'
+
+    deprecated_directive_major_version:
+        major_version: 3
+
+    deprecated_directive_min_version:
+        min_version: '3.0'
```
OskarStark: After up merging into 4.x this must be change accordingly. Same for `master`, if this will be next 5.0, otherwise not.

OskarStark:
> I will disable the rule and revert the commit

done

OskarStark:
For the record we use this extension for symfony-docs Sphinx:
https://github.com/fabpot/sphinx-php#sphinx-extensions-for-php-and-symfony

malarzm:
Let's get this merged, thanks a lot @OskarStark!

OskarStark:
Wohoo 🙌 thanks 🥳

SenseException:
Thank you :tulip:

PR #614, opened 2020-02-11
Make manager registry resettable

Baldinof:
Hi!

This is shamely copy/pasted from [Doctrine ORM Bundle](https://github.com/doctrine/DoctrineBundle/blob/master/Registry.php) 😅


alcaeus:
With all the problems that were caused by doing this in DoctrineBundle, I'm not so sure I want to do this again. 🤔

Baldinof:
As you want! Aren't the problems fixed in the current implementation?

PR #615, opened 2020-02-11
Incorrect ManagerRegistry.

tlorens:
Error is thrown: Could not find the document manager for class '...' Check your Doctrine configuration to make sure it is configured to load this document’s metadata.

### DIFF
--- a/Resources/doc/first_steps.rst
+++ b/Resources/doc/first_steps.rst
@@ -452,7 +452,7 @@ is to use the repository as a service and inject it as a dependency into other s
 
     use Acme\StoreBundle\Document\Product;
     use Doctrine\Bundle\MongoDBBundle\Repository\ServiceDocumentRepository;
-    use Doctrine\Common\Persistence\ManagerRegistry;
+    use Doctrine\Bundle\MongoDBBundle\ManagerRegistry;
 
     /**
      * Remember to map this repository in the corresponding document's repositoryClass.


Resources/doc/first_steps.rst
```diff
@@ -452,7 +452,7 @@ is to use the repository as a service and inject it as a dependency into other s
 
     use Acme\StoreBundle\Document\Product;
     use Doctrine\Bundle\MongoDBBundle\Repository\ServiceDocumentRepository;
-    use Doctrine\Common\Persistence\ManagerRegistry;
+    use Doctrine\Bundle\MongoDBBundle\ManagerRegistry;
```
SenseException: Thanks for the fix. I assume it should still be the `ManagerRegistry` if the Persistence library.
```suggestion
    use Doctrine\Persistence\ManagerRegistry;
```

tlorens: I don't think so, I pasted the code I used while working with ODM/MongoDB.   I've have since moved on from that solution/prototype.  I could dig back into it maybe.  I'd have to find it.

tlorens: My MongoDB 'entity' repository class.  The example in the documentation throws an error as I stated before. 

`<?php

namespace App\Repository;

use App\Document\Audit;
use Doctrine\Bundle\MongoDBBundle\Repository\ServiceDocumentRepository;
use Doctrine\Bundle\MongoDBBundle\ManagerRegistry;

/**
 * @method Audit|null find($id, $lockMode = null, $lockVersion = null)
 * @method Audit|null findOneBy(array $criteria, array $orderBy = null)
 * @method Audit[]    findAll()
 * @method Audit[]    findBy(array $criteria, array $orderBy = null, $limit = null, $offset = null)
 */
class AuditDocumentRepository extends ServiceDocumentRepository
{
    public function __construct(ManagerRegistry $registry)
    {
        parent::__construct($registry, Audit::class);
    }

    public function save(Audit $audit)
    {
        $this->_em->persist($audit);
        $this->_em->flush();
    }
}
`

malarzm:
@tlorens it seems this issue was fixed along with documentation improvements that were applied recently thus I'm closing the PR. Thank you for your time and letting us know something was off!

PR #617, opened 2020-02-14
Address deprecations from persistence library

### LINKED ISSUE #616: Incompatibility with Symfony 4.3.10+
Hello

In our company's project we are currently using Symfony 4.4 branch, but we are still behind in our move towards using Doctrine ODM 2.

Therefore we still rely on doctrine/mongodb-odm-bundle for the interface beween Symfony and doctrine odm 1.3.

Unfortunately, the PR https://github.com/symfony/symfony/pull/34955 has broken interface compatibility between the doctrine/mongodb-odm-bundle and Symfony's latest versions (4.4.2+,4.3.10+,5.0.2+):

```
PHP Fatal error:  Declaration of Doctrine\Bundle\MongoDBBundle\Form\Type\DocumentType::getLoader(Doctrine\Common\Persistence\ObjectManager $manager, $queryBuilder, $class) 
must be compatible with Symfony\Bridge\Doctrine\Form\Type\DoctrineType::getLoader(Doctrine\Persistence\ObjectManager $manager, $queryBuilder, $class)
```

Since the interface is almost the same, I think the fix should be just changing the method `Doctrine\Bundle\MongoDBBundle\Form\Type\DocumentType::getLoader` to receive a `Doctrine\Persistence\ObjectManager` but I'm not sure if this would cause any other incompatibilities.

greg0ire:
A backwards-compatibility layer has been added to persistence to help
consumers move to the new namespacing. It is based on class aliases,
which means the type declaration changes should not be a BC-break: types
are the same.
See doctrine/persistence#71

This means:
- using the new namespaces
- adding autoload calls for new types to types that may be extended and
use persistence types in methods signatures for which compatibility
checks may happen,
so that signature compatibility is recognized by old versions of php.
More details on this at
https://dev.to/greg0ire/how-to-deprecate-a-type-in-php-48cf

Fixes #616 as a side effect, but the real fix is at
https://github.com/symfony/symfony/pull/35728


greg0ire:
Is it the correct target branch? Do we consider "fixing" deprecations a patch? Note that it's future deprecations…

greg0ire:
Good catch, fixed! And no it shouldn't fail IMO, the old name is an alias for the new one, so using one in place of the other should produce the same result.

alcaeus:
> Is it the correct target branch? Do we consider "fixing" deprecations a patch? Note that it's future deprecations…

Sorry for the long delay here. Yes, we will fix these in a patch release, and yes, this is the right target branch: 3.6 can't deal with the new persistence version as it must be compatible with PHP 5, so 4.0 is the lowest version that gets these fixed.

alcaeus:
Ah, small correction: 4.1.x is the current supported minor version. Could you please rebase to fix conflicts? Thanks!

greg0ire:
@alcaeus done :)

malarzm:
Thanks a lot @greg0ire!

PR #631, opened 2020-05-24
Support custom types in configuration

shadz3rg:
Currently no problems with DM usage and doctrine:mongodb:generate:hydrators command.
It is possible to move method `\Doctrine\Bundle\MongoDBBundle\ManagerConfigurator::loadTypes` in separate service, but it is kinda looking OK in abstract Configurator.


g13013:
it should be mentioned in the documentation that types keys are converted to snake_case so `some-type` will be registered as `some_type`

PR #633, opened 2020-05-29
[Messenger] Clear DocumentManagers after a message is handled or failed

smoench:
This PR registers the [DoctrineClearEntityManagerWorkerSubscriber](https://github.com/symfony/doctrine-bridge/blob/master/Messenger/DoctrineClearEntityManagerWorkerSubscriber.php) which clears the document managers after each message is handled. The classname is misleading but it works with any ManagerRegistry which implements the `Doctrine\Persistence\ManagerRegistry`.

I'm using it in production and it works fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ before_install:
 install:
   - composer self-update
   - pecl install -f mongodb-${DRIVER_VERSION}
+  - if [[ -z ${REQUIRE_SYMFONY_MESSENGER} ]]; then composer require --dev symfony/messenger --no-update; fi
   - composer update ${COMPOSER_FLAGS}
 
 script:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -26,6 +26,7 @@ before_install:
 install:
   - composer self-update
   - pecl install -f mongodb-${DRIVER_VERSION}
+  - if [[ -z ${REQUIRE_SYMFONY_MESSENGER} ]]; then composer require --dev symfony/messenger --no-update; fi
```
malarzm, line 29: Could we use the same approach to installing `symfony/messenger` as DoctrineBundle? https://github.com/doctrine/DoctrineBundle/blob/a9e3720814d93789bd7c91e80957fcd1003bf3b6/.travis.yml#L52

malarzm (commented):
Thanks for the PR! Would you mind changing travis configuration?

.travis.yml
smoench: I tried it before but I had to copy all commands to each include item as the install section gets overridden. I would prefer my approach as it results in less duplicated lines.

malarzm:
Thanks a lot @smoench!

PR #637, opened 2020-06-03
Add section for Messenger integration

### LINKED ISSUE #634: Add documentation for messenger integration
That was implemented in #633

malarzm:
Closes #634

### DIFF
--- a/Resources/doc/messenger.rst
+++ b/Resources/doc/messenger.rst
@@ -1,0 +1,8 @@
+Messenger Integration
+=====================
+
+When using Symfony 4.4 or later, the bundle automatically registers a `Messenger`_
+event subscriber that clears all document managers after handling messages, which
+helps to isolate each handler and guard against reading out-of-date document data.
+
+.. _`Messenger`: https://symfony.com/doc/current/components/messenger.html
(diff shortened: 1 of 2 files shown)


Resources/doc/messenger.rst
```diff
@@ -0,0 +1,11 @@
+Messenger integration
+=====================
+
+The bundle automatically registers a `Messenger`_ middleware that clears
+document managers between messages being handled to avoid outdated data.
```
jmikola, line 5: Suggestion:

> ...that clears all document managers after handling messages, which helps to isolate each handler and guard against reading out-of-date document data.

I summarized that from @weaverryan's explanation in https://github.com/symfony/symfony/pull/34156. You could also considering adding a link to the issue for more context, since there isn't much else to say on this page, but I don't feel strongly either way about that.

malarzm: Dang, now I see it's not even a middleware but rather a event subscriber 👀 I'll go with your summary

jmikola: Looking at the Symfony PR, I think the design changed somewhere along the way. In any event, current text LGTM.

PR #638, opened 2020-06-03
Section for custom types in configuration chapter

### LINKED ISSUE #632: Add documentation for custom types
Feature was added in https://github.com/doctrine/DoctrineMongoDBBundle/pull/631

malarzm:
Closes #632

### DIFF
--- a/Resources/doc/config.rst
+++ b/Resources/doc/config.rst
@@ -214,6 +214,36 @@ The following configuration shows a bunch of mapping examples:
             </doctrine_mongodb:config>
         </container>
 
+Custom Types
+------------
+
+`Custom types`_ can come in handy when you're missing a specific mapping type
+or when you want to replace the existing implementation of a mapping type for
+your documents.
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        doctrine_mongodb:
+            types:
+                custom_type: Fully\Qualified\Class\Name
+
+    .. code-block:: xml
+
+        <?xml version="1.0" ?>
+
+        <container xmlns="http://symfony.com/schema/dic/services"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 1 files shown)


Resources/doc/config.rst
```diff
@@ -214,6 +214,39 @@ The following configuration shows a bunch of mapping examples:
             </doctrine_mongodb:config>
         </container>
 
+Custom types
```
jmikola, line 217: Capitalize "Types" since this is a title

Resources/doc/config.rst
```diff
@@ -214,6 +214,39 @@ The following configuration shows a bunch of mapping examples:
...
+or when you want to replace the existing implementation of a mapping type for
+your documents.
+
+.. configuration-block::
+
+    .. code-block:: yaml
+
+        doctrine_mongodb:
+            types:
+                # following two definitions are equivalent
```
jmikola, line 230: If I understand correctly, you're saying that the value can either be a FQCN string directly or an object with the FQCN string under a `class` property. I think it may be simpler to just present one style in the documentation example, though, unless there is a legitimate reason users would need to use the more verbose syntax (perhaps if other options beyond `class` were expected).

Resources/doc/config.rst
```diff
@@ -536,6 +569,10 @@ Full Default Configuration
                             prefix:               ~
                             alias:                ~
                             is_bundle:            ~
+            types:
+
+                # Prototype
+                name: Fully\Qualified\Class\Name
```
jmikola, line 575: I'd suggest using "custom_type" instead of "name" here, for consistency with the previous example. Likewise in the XML below.

Resources/doc/config.rst
malarzm: Not right now, indeed it'll be better to just show simpler syntax 👍

PR #649, opened 2020-09-08
Update load fixtures command description

franmomu:
This fixes the tag service name and removes reference to some nonexistent options.


malarzm:
@franmomu thanks for the PR! It looks like we missed the fact we wanted the option gone so I've created #650 to track this. Your changes look good and I'll merge once I figure out what happened with travis job :)

PR #652, opened 2020-10-02
Improve collector view

### LINKED ISSUE #319: Show query time in profiler
In dev mode, it would be great to see the query time for each query (like the Doctrine ORM profiler does: http://take.ms/37ILV). This makes it a lot easier to identify queries that are subject for performance improvement.

Jean-Beru:
This PR makes the collector show execution time of commands. Style has been taken from DoctrineBundle and could be improved in future PR (eg: a better command display).

We may add a better command display later. Eg: `db.User.find({ _id: 1 });`.
image


Resources/views/Collector/mongodb.html.twig
```diff
@@ -34,21 +34,43 @@
...
+            <th style="width: 100%;">Info</th>
+        </tr>
+        </thead>
+        <tbody id="commands">
+        {% for command in collector.commands %}
+            <tr id="queryNo-{{ loop.index0 }}">
+                <td class="nowrap">{{ loop.index }}</td>
+                <td class="nowrap">{{ command.durationMicros }} ms</td>
+                <td>
+                    {{ command.command }}
```
alcaeus: Should be wrapped in `<code>` blocks:
```suggestion
                    <code>{{ command.command }}</code>
```

Resources/views/Collector/mongodb.html.twig
```diff
@@ -34,21 +34,43 @@
...
+        <thead>
+        <tr>
+            <th class="nowrap">#</th>
+            <th class="nowrap">Time</th>
+            <th style="width: 100%;">Info</th>
+        </tr>
+        </thead>
+        <tbody id="commands">
+        {% for command in collector.commands %}
+            <tr id="queryNo-{{ loop.index0 }}">
```
alcaeus: Strictly speaking, we're not doing queries, so `command-...` seems more appropriate:
```suggestion
            <tr id="command-{{ loop.index0 }}">
```

Resources/views/Collector/mongodb.html.twig
```diff
@@ -34,21 +34,43 @@
...
-                        <code>{{ command }}</code>
-                    </div>
-                </li>
-            {% endfor %}
-        </ul>
+        <div class="empty">
+            <em>No commands were performed.</em>
+        </div>
     {% endif %}
+    <table class="alt queries-table">
```
alcaeus: Does `queries-table` have any styling associated with it by default?

Jean-Beru: Nope. Just a bad copy/paste :sweat:

DataCollector/CommandDataCollector.php
```diff
@@ -29,11 +30,21 @@ public function collect(Request $request, Response $response, ?Throwable $except
         $this->data = [
             'num_commands' => count($this->commandLogger),
             'commands' => array_map(
-                static function (Command $command) : string {
-                    return json_encode($command->getCommand());
+                static function (Command $command) : array {
```
alcaeus: While writing a comment to suggest tests, I realised that the current test calls `collect`, but does not perform any assertion. May I ask you to add the assertions for the `getTime` method you added here? Feel free to ignore the command lists, but if we could have some test for the time logic, that would be great. Please let me know if this works for you.

Jean-Beru: I tried to add tests for `CommandDataCollector` but `Command`, `CommandStartedEvent` or `CommandSucceededEvent` are final and have private properties. So, I don't know how to make tests work since we cannot instantiate or mock them.

Jean-Beru: Maybe we can consider to use dg/bypass-finals. WDYT ?

Jean-Beru: @alcaeus I added https://github.com/dg/bypass-finals to allow APM mocks.Is it OKay for you ?

Jean-Beru:
Is there anything that I can do to make this PR OK ?

alcaeus:
I don't have enough time to review things at the moment, so there isn't much you can do. I'll mark this so the bot doesn't keep reminding you that I don't have time 😞

Jean-Beru:
I totally understand. I thought that you were waiting for some changes. Good luck and happy holiday season :wink:

DataCollector/CommandDataCollector.php
```diff
@@ -29,11 +30,21 @@ public function collect(Request $request, Response $response, ?Throwable $except
...
+                    return [
+                        'command' => json_encode($command->getCommand()),
```
IonBazan: Maybe we can use `$this->cloneVar()` to create a vardumper clone and perform json encoding on the template instead? We could add the raw dump (`profiler_dump()`) of the command there and pretty-print the JSON too.

Resources/views/Collector/mongodb.html.twig
```diff
@@ -34,21 +34,43 @@
...
+        <div class="empty">
+            <em>No commands were performed.</em>
+        </div>
     {% endif %}
+    <table>
+        <thead>
+        <tr>
+            <th class="nowrap">#</th>
+            <th class="nowrap">Time</th>
+            <th style="width: 100%;">Info</th>
```
IonBazan, line 62: Let's add the DB name too.

IonBazan (commented):
@alcaeus This feature would quite useful to many people. I was actually starting working on something similar but then found this PR. Is there any chance you could find a moment to take a look on that again? 🙏🏻

Resources/views/Collector/mongodb.html.twig
Jean-Beru: Good idea !

DataCollector/CommandDataCollector.php
Jean-Beru: Using `profiler_dump` will add a dependency to WebProfilerBundle. So, I just moved JSON encoding to template.

Maybe we could add a MongoDB query formatter later in order to display commands like this : `db.User.find({ _id: 1 });`

noniagriconomie:
May I ask you some screenshots? I would definitly love to have this improved UI as well, thanks :)

Jean-Beru:
Of course, I have just updated PR's description.

DataCollector/CommandDataCollector.php
IonBazan: @Jean-Beru I don't think adding a dependency is necessary for that. [DoctrineBundle](https://github.com/doctrine/DoctrineBundle/blob/2.4.x/composer.json) references `symfony/web-profiler-bundle` in `require-dev` and `suggest` only. This is how most bundles that use DataCollectors behave.

That would be way cleaner than json-encoding the value to persist it in the profiler storage.

DataCollector/CommandDataCollector.php
alcaeus: Sorry for the long delay. I've taken a look, and I don't think it's necessary to mock this. We don't need to check for the exact time, but inserting a document should get some commands logged so we can assert that the time is greater than 0 for all commands. Please let me know if you can make the changes or if you want me to take this over. Thanks for your work so far, this looks good apart from the testing!

DataCollector/CommandDataCollector.php
alcaeus: @Jean-Beru we won't be able to format these commands into anything other than `db.runCommand(...)`, which is why I haven't done so. Furthermore, all cursor-involving commands are not retryable, so we shouldn't encourage users to actually run these commands.

As for the output, `json_encode` is fine, but since MongoDB uses BSON for its command, it may be more appropriate to change this to:
```php
use function MongoDB\BSON\fromPHP;
use function MongoDB\BSON\toCanonicalExtendedJSON;
// ...
return [
    'command' => toCanonicalExtendedJSON(fromPHP($command->getCommand()),
];
```

Since performance is not an issue, we don't need to worry about converting the PHP object back into a BSON string and then converting that to our [Extended JSON format](https://github.com/mongodb/specifications/blob/master/source/extended-json.rst). This preserves type information and is understood by all MongoDB drivers.

For now, I'm fine leaving the current output format in place, but this could be an improvement for future versions to let users choose whether they want plain json or extended json for use in other drivers.

DataCollector/CommandDataCollector.php
Jean-Beru: End of my holidays, I have updated tests to remove mocks !

DataCollector/CommandDataCollector.php
Jean-Beru: @alcaeus I added extended JSON format. Maybe we have to add `ext-mongodb` to requirements since these functions depend on `MongoDB\BSON` namespace ?

alcaeus: Yep, we should. Depending on `ext-mongodb: ^1.0` would be the right constraint here.

Jean-Beru: Done :wink:

composer.json
```diff
@@ -13,6 +13,7 @@
     ],
     "require": {
         "php": "^7.2 || ^8.0",
+        "ext-mongodb": "^1.0",
```
IonBazan: I don't think it's actually needed here. This package depends on `doctrine/mongodb-odm ^2.0` which requires `ext-mongodb ^1.5` already so this requirement is superfluous.

Jean-Beru: Since we use `MongoDB\BSON\fromPHP` and `MongoDB\BSON\toCanonicalExtendedJSON`, I think we should define this dependency explicitly.
But, you're right, we can set it to `^1.5`.

noniagriconomie (approved):

IonBazan (approved):
✨

composer.json
alcaeus: Picking the lowest version that contains the feature set you want makes sense. Adding the constraint is necessary, but 1.0 would've done. I'm not going to ask you to change it back though

composer.json
```diff
@@ -9,10 +9,11 @@
     "authors": [
         {"name": "Bulat Shakirzyanov", "email": "mallluhuct@gmail.com"},
         {"name": "Kris Wallsmith", "email": "kris@symfony.com"},
-        {"name": "Jonathan H. Wage", "email": "jonwage@gmail.com"}
+        {"name": "Jonathan H. Wage", "em    ail": "jonwage@gmail.com"}
```
alcaeus: This looks like an unwanted change 😉

alcaeus (commented):
LGTM with the email change reverted 😊

composer.json
Jean-Beru: :man_facepalming:

alcaeus (approved):

Resources/views/Collector/mongodb.html.twig
```diff
@@ -34,21 +34,43 @@
...
+        </tr>
+        </thead>
+        <tbody id="commands">
+        {% for command in collector.commands %}
+            <tr>
+                <td class="nowrap">{{ loop.index }}</td>
+                <td class="nowrap">{{ '%0.3f'|format(command.durationMicros / 1000) }} ms</td>
+                <td class="nowrap">{{ command.database }}</td>
+                <td>{{ command.command }}</td>
```
IonBazan: Raw JSONs are hard to read so I think we could add a collapsible pretty-print version of the command JSON here just like in ElasticaBundle: https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/src/Resources/views/Collector/elastica.html.twig#L129-L135

Jean-Beru: What about decoding and encoding current value in template to keep canonical extended JSON format ? Like `{{ data|json_decode|json_encode(constant('JSON_PRETTY_PRINT')) }}`.

IonBazan: Yes, that makes sense 👍🏻

Jean-Beru: Done. FYI, `json_decode` has been done in DataCollector since Twig does not provide a such filter, cf. https://github.com/twigphp/Twig/pull/824.

Resources/views/Collector/mongodb.html.twig
```diff
@@ -68,7 +68,9 @@
                 <td class="nowrap">{{ loop.index }}</td>
                 <td class="nowrap">{{ '%0.3f'|format(command.durationMicros / 1000) }} ms</td>
                 <td class="nowrap">{{ command.database }}</td>
-                <td>{{ command.command }}</td>
+                <td>
+                    <pre>{{ command.command|json_encode(constant('JSON_PRETTY_PRINT')) }}</pre>
```
IonBazan: I think we should show the non-pretty-print one first and allow to expand it for the pretty print one. You can use `data-toggle-selector` attributes to achieve it. See https://github.com/doctrine/DoctrineBundle/blob/2.4.x/Resources/views/Collector/db.html.twig

Jean-Beru: Done by using existing `queries-table` and `sql-runnable` CSS classes.

Jean-Beru:
The CI failed :
```
==> Setup Extensions
✗ mongodb-1.10.12 Could not install mongodb-1.10.12 on PHP 7.2.34
Error: The process '/usr/bin/bash' failed with exit code 1
```
Is it due to `ext-mongodb` in `composer.json` ? :thinking:

alcaeus:
I'll have to investigate what's going on here: 1.10.12 does not exist, and the build specifically calls for installing 1.5.0. For some reason, the version matches what we had above for pecl, so I'm inclined to think there's a separate issue going on. I'll take a look.

alcaeus:
I've updated the target branch to 4.4.x and rebased, which of course adds psalm errors. @franmomu would you be able to add the array shapes to this so we can wrap it up? Thanks!

franmomu:
There you go!

franmomu (approved):

alcaeus:
You're a legend, @franmomu! Thanks @Jean-Beru and @IonBazan for getting this one done! 🎉

PR #653, opened 2020-10-21
Bump to PHPUnit 8.5 and tests clean up

franmomu:
I was trying to add psalm and there were some issues in `Tests`, so this is a bit of clean up Tests before adding psalm. 

It is split in commits, the relevant ones:

- Psalm was complaining about some non-existing classes, I created some dummy ones.
- Update tests to be compatible with PHPUnit 8 and 9.
- ~[Drop support for doctrine/data-fixtures < 1.3, I could do this in a separate PR if it's necessary.](https://github.com/doctrine/DoctrineMongoDBBundle/pull/655)~


franmomu:
Just commenting to have some activity.

franmomu:
~Where did travis go? 🤔~ https://github.com/doctrine/DoctrineMongoDBBundle/pull/660

Tests/Mapping/Driver/AbstractDriverTest.php
```diff
@@ -37,25 +38,23 @@ public function testFindMappingFileInSubnamespace()
...
     }
 
-    /**
-     * @expectedException Doctrine\Persistence\Mapping\MappingException
-     */
     public function testFindMappingFileNamespacedFoundFileNotFound()
     {
         $driver = $this->getDriver([$this->getFixtureDir() => 'MyNamespace\MyBundle\Document']);
 
+        $this->expectException(MappingException::class);
```
greg0ire: This line should be moved right before the one that triggers the exception (so probably the last one, right?)

greg0ire (commented):
Very nice, just one tiny nitpick.

malarzm (approved):

malarzm:
Thanks @franmomu!

PR #658, opened 2020-11-12
fix warmup persistent collection

panayotovyura:
Hi, I have discovered that the configuration flag `auto_generate_persistent_collection_classes` doesn't work as it supposed to. Pls, check if I'm right.


alcaeus:
You don't want to return from the warmup method if the mode is `AUTOGENERATE_NEVER`, as that will cause runtime failures. We want to return if autogeneration is enabled, which is when mode is unequal to `AUTOGENERATE_NEVER`.

panayotovyura:
Hm, but right now I get runtime errors when I don't have autogenerated classes... I expect to set `auto_generate_persistent_collection_classes` to AUTOGENERATE_FILE_NOT_EXISTS or AUTOGENERATE_ALWAYS and symfony command `php bin/console cache:warmup` will generate those files for me. But right now it works vice versa

PR #662, opened 2020-12-08
Add support for PHP 8

lepiaf:


lepiaf:
waiting for merge of PR https://github.com/doctrine/mongodb-odm/pull/2248

malarzm:
@lepiaf thanks for the initial work! This will need to wait until we release 2.2, but in the meantime maybe you could change ODM to require @alcaeus branch if it's in green-ish state? That could allow you to check if any further work is required

franmomu:
So there has been a release: https://github.com/doctrine/mongodb-odm/releases/tag/2.2.0 😬

franmomu:
Just in case, I created yesterday https://github.com/doctrine/DoctrineMongoDBBundle/pull/666.

lepiaf:
I have this error in phpunit for PHP 8.
```
Generating code coverage report in Clover XML format ... failed [00:01.648]
Cannot parse /home/runner/work/DoctrineMongoDBBundle/DoctrineMongoDBBundle/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/Aggregation/Stage/GraphLookup/Match.php: Syntax error, unexpected T_MATCH, expecting T_STRING on line 15
```

In PHP 8, `Match` is a reserved keyword. This error comes from `doctrine/mongodb-odm` library. There is already a deprecation error on it https://github.com/doctrine/mongodb-odm/blob/2.2.0/UPGRADE-2.2.md#aggregation.

This error is raised on code coverage.

Tests/DataCollector/CommandDataCollectorTest.php
```diff
@@ -17,5 +17,6 @@ public function testCollector()
         $collector = new CommandDataCollector(new CommandLogger());
 
         $collector->collect($request = new Request(['group' => '0']), $response = new Response());
+        $this->assertSame(0, $collector->getCommandCount());
```
lepiaf, line 23: test was marked as risky. so I've added at least one assert.

composer.json
```diff
@@ -29,7 +29,7 @@
         "doctrine/data-fixtures": "<1.3"
     },
     "require-dev": {
-        "doctrine/coding-standard": "^6.0",
+        "doctrine/coding-standard": "^8.0",
```
lepiaf: I let https://github.com/doctrine/DoctrineMongoDBBundle/pull/666 fix all coding style

Tests/Fixtures/Form/Guesser.php
```diff
@@ -35,18 +35,12 @@ class Guesser
     /** @ODM\Field(type="bool") */
     public $boolField;
 
-    /** @ODM\Field(type="boolean") */
```
lepiaf, line 38: I had this error in phpunit
```
Remaining indirect deprecation notices (2)

  1x: Since doctrine/mongodb-odm 2.1: The "boolean" mapping type is deprecated. Use "bool" instead.
    1x in TypeGuesserTest::testTypesShouldBeGuessedCorrectly from Doctrine\Bundle\MongoDBBundle\Tests\Form\Type

  1x: Since doctrine/mongodb-odm 2.1: The "integer" mapping type is deprecated. Use "int" instead.
    1x in TypeGuesserTest::testTypesShouldBeGuessedCorrectly from Doctrine\Bundle\MongoDBBundle\Tests\Form\Type
```

composer.json
franmomu: #666 was merged, I think there is no need to modify those tests anymore.

malarzm (approved):

Tests/Fixtures/Form/Guesser.php
malarzm: It'd be the best to extract these two asserts into a separate test that would allow deprecation notices but considering simplicity of the guesser I think we'll be just fine

malarzm:
Thanks a lot @lepiaf!

PR #665, opened 2021-01-03
PHP8 support

rimvislt:


rimvislt:
Not sure how to fix coding standards checkout issue

alexislefebvre:
#662 and this issue are duplicates.

composer.json
```diff
@@ -12,15 +12,15 @@
         {"name": "Jonathan H. Wage", "email": "jonwage@gmail.com"}
     ],
     "require": {
-        "php": "^7.2",
-        "doctrine/annotations": "^1.2",
-        "doctrine/mongodb-odm": "^2.0.0",
+        "php": "^7.2 || ^8.0",
```
dbrumann, line 15: Would it make sense to use `>=7.2` instead to keep it in line with Symfony's preferred versioning strategy?

composer.json
```diff
@@ -12,15 +12,15 @@
...
+        "doctrine/annotations": "^1.6",
+        "doctrine/mongodb-odm": "^2.2",
```
dbrumann: Is this really necessary? The old `^2.0.0` would still allow 2.2 for PHP 8, but right now someone with PHP 7.2 would no longer be able to use doctrine/mongodb-odm 2.0/2.1.

composer.json
malarzm: No, that's a terrible strategy

composer.json
malarzm: Indeed, the `^2.0.0` constraint should stay (unless there are some blockers but I don't think there'll be any)

rimvislt: Thank you guys for feedback reverted back to 2.0.0

rimvislt:
Should I update code to minimum php 7.3, because php 7.2 is end of live? (https://www.php.net/supported-versions.php)

dbrumann:
I don't think updating the minimum PHP version should be part of this PR. Since Symfony 5 supports PHP 7.2 it makes sense to keep it unless there is something to be gained from dropping it.

malarzm:
> Should I update code to minimum php 7.3, because php 7.2 is end of live? (https://www.php.net/supported-versions.php)

The best for the bundle is to allow same PHP versions as ODM does: let's keep 7.2 for now. If we are going to drop it we will need to release another ODM version first.

malarzm:
I'm going to close the PR as I've just merged #662. That PR got chosen because it had all checks passing already. Nevertheless I want to thank you @rimvislt for your work on the issue! :)

PR #659, opened 2020-11-24
Address master rename

greg0ire:

PR #660, opened 2020-11-29
Move tests to GitHub Actions

franmomu:
This is based on https://github.com/doctrine/mongodb-odm/pull/2250

I've changed to check just Symfony maintained versions 4.4 and 5.1.

Some doubts:

I don't know if `composer global require --no-progress --no-scripts --no-plugins symfony/flex dev-master` is necessary.
And also about this:
```yml
    # Test dev versions
    - php: 7.3
      if: type = cron
      env: DEV COMPOSER_FLAGS="-n --prefer-dist"
```

The other cases I think they are covered.

PR #666, opened 2021-01-24
Update coding standards to 8.2

franmomu:
This tries to use `doctrine/coding-standard` 8, to be able to allow php 8.

Tests should pass after merging https://github.com/doctrine/DoctrineMongoDBBundle/pull/659 that fixes some deprecations.

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -39,7 +39,7 @@ jobs:
           restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
 
       - name: "Install dependencies with Composer"
-        run: "composer install --no-interaction --no-progress --no-suggest"
+        run: "composer update --no-interaction --no-progress --prefer-dist --prefer-stable"
 
       # https://github.com/doctrine/.github/issues/3
       - name: "Run PHP_CodeSniffer"
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -31,15 +31,11 @@ jobs:
           php-version: "${{ matrix.php-version }}"
           tools: "cs2pr"
 
-      - name: "Cache dependencies installed with Composer"
-        uses: "actions/cache@v2"
-        with:
-          path: "~/.composer/cache"
-          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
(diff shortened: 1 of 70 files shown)


franmomu:
I cherry-picked the commits (I'll remove them if necessary), it always nice to see it green 🟢

SenseException:
@greg0ire The cherry-picked commits might cause problems when both branches get merged. If that's the case I suggest to remove the commits and rebase later when your changes are in the base branch.

.github/workflows/coding-standards.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
 
       - name: "Install dependencies with Composer"
-        run: "composer install --no-interaction --no-progress --no-suggest"
+        run: "composer update --no-interaction --no-progress --prefer-dist --prefer-stable"
```
greg0ire: This file seems quite a bit out of sync with [upstream](https://github.com/doctrine/.github/blob/main/workflow-templates/coding-standards.yml), let's use the same solution instead?

greg0ire:
> The cherry-picked commits might cause problems when both branches get merged.

What kind of problems? In my experience cherry-picking does not usually cause issues.

.github/workflows/coding-standards.yml
franmomu: Now I remembered that I tried, getting this:

https://github.com/doctrine/DoctrineMongoDBBundle/runs/1787972990?check_suite_focus=true

As the minimum stability is `dev`, it installs [`dev-master` of `slevomat/coding-standard`](https://github.com/doctrine/DoctrineMongoDBBundle/runs/1787972990?check_suite_focus=true#step:4:69) which removes [`SlevomatCodingStandard.Classes.UnusedPrivateElements`](https://github.com/slevomat/coding-standard/commit/655524a1a70b6b1cc0f75bb5c1b7c6b8286dad86).

The `dev` stability was added in https://github.com/doctrine/DoctrineMongoDBBundle/pull/602 to apparently test dev versions in travis and it was not added in #660, so maybe a solution could be to remove this `dev` stability.

greg0ire: Yes, adding a step that does `composer config minimum-stability stable` could be a solution.

greg0ire (approved):

SenseException (approved):

PR #671, opened 2021-03-02
Use proper config attribute in XML mapping

noniagriconomie:
Fix this code snippet
Reference to: `symfony console doctrine:mongodb:mapping:info`

The mapping file [...]/src/Resources/config/doctrine/Item.mongodb.xml is invalid: 
Line 7:0: Element '{http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping}field', attribute 'fieldName': The attribute 'fieldName' is not allowed.


malarzm:
Thanks @noniagriconomie!

PR #681, opened 2021-07-18
Remove deprecated code

franmomu:
`FormTypeInterface::getBlockPrefix` was added in symfony/symfony 3.0, see https://github.com/symfony/symfony/blob/3.0/UPGRADE-3.0.md#form and we dropped support for 3.4 some time ago.

I've added another commit adding return types.


malarzm (approved):

malarzm:
Thanks @franmomu!

PR #683, opened 2021-07-20
Make cache warmers compatible with symfony 5.1

franmomu:
Since version 5.1, it must return an array with the files and classes to preload, if there is no need to preload an empty array should be returned.

Another issue that came up when trying to raise psalm level (using [psalm/psalm-plugin-symfony](https://github.com/psalm/psalm-plugin-symfony)) is that the whole container is injected to cache warmers, so should we mark them as `internal` for the next major version? I don't think they are supposed to be used directly by users, that way they would be out of the BC promise.


alcaeus (approved):

alcaeus:
Thanks @franmomu!

malarzm:
@franmomu just to answer this part:

> Another issue that came up when trying to raise psalm level (using psalm/psalm-plugin-symfony) is that the whole container is injected to cache warmers, so should we mark them as internal for the next major version? I don't think they are supposed to be used directly by users, that way they would be out of the BC promise.

In ODM (https://github.com/doctrine/mongodb-odm/pull/2008) we went through all classes that we considered non-extendable ones, marked them as `@final` and make them emit a deprecation notice if anybody extended them. Later in 2.0 they became true final classes. In addition a class might be additionally marked as `@internal`. Sadly the bundle hasn't seen such action but it might be worthwhile to do so :)

PR #675, opened 2021-05-19
Add Configless Metadata Cache

alcaeus:
This PR deprecates the `metadata_cache_driver` config key in favour of automatic configuration. In debug mode, it creates an ArrayAdapter, meaning that the cache is only valid per request. This is usually desired in a development environment. When debug is disabled, it creates a PhpArrayAdapter, backed by an array cache. The cache is warmed up and stored in the file system. Since this information shouldn't be shared between multiple instances of the code, it is only logical to store this in the cache folder in an efficient format. This also mirrors a decision made in the ORM bundle to deprecate the config.

Note: This PR contains a commit from #674. I'll rebase this PR once the other one has been merged.

PR #685, opened 2021-07-25
Add configless metadata caching

Ahummeling:
Quoting @alcaeus description of the related PR #675 
> This PR deprecates the metadata_cache_driver config key in favour of automatic configuration. In debug mode, it creates an ArrayAdapter, meaning that the cache is only valid per request. This is usually desired in a development environment. When debug is disabled, it creates a PhpArrayAdapter, backed by an array cache. The cache is warmed up and stored in the file system. Since this information shouldn't be shared between multiple instances of the code, it is only logical to store this in the cache folder in an efficient format. This also mirrors a decision made in the ORM bundle to deprecate the config.


DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -496,4 +502,71 @@ public function getXsdValidationBasePath()
...
+        if (is_a($cacheDefinition->getClass(), CacheItemPoolInterface::class, true)) {
+            return;
+        }
+
+        $adapterDefinition = new Definition(CacheItemPoolInterface::class);
+        $adapterDefinition
+            ->setFactory([CacheAdapter::class, 'wrap'])
+            ->setArgument(0, $cacheDefinition);
+
+        $container->setDefinition($cacheServiceId, $adapterDefinition);
```
alcaeus, line 547: This part won't work like that, as we can't check the cache definitions class at this stage. In DoctrineBundle, we worked around this by adding a compiler pass that converts caches to the correct instance: https://github.com/doctrine/DoctrineBundle/blob/2.4.x/DependencyInjection/Compiler/CacheCompatibilityPass.php

ODM only has metadata caching, so we don't need to account for other caches like we do for ORM, but the logic can stay the same.

In the extension, we then set up the metadata cache layer the way it has been, with the compiler pass ensuring that the correct cache instance is passed to the configuration.

Ahummeling:
Implementing psr6 compatibility pass in [this pr](https://github.com/doctrine/DoctrineMongoDBBundle/pull/690)

PR #687, opened 2021-07-28
Bump psalm to level 4

franmomu:
I'll add some comment in the code to explain some changes.

### DIFF
--- a/CacheWarmer/HydratorCacheWarmer.php
+++ b/CacheWarmer/HydratorCacheWarmer.php
@@ -23,6 +23,10 @@ use function sprintf;
  *
  * In the process of generating hydrators the cache for all the metadata is primed also,
  * since this information is necessary to build the hydrators in the first place.
+ *
+ * @internal since version 4.4
+ *
+ * @psalm-suppress ContainerDependency
  */
 class HydratorCacheWarmer implements CacheWarmerInterface
 {
--- a/CacheWarmer/HydratorCacheWarmer.php
+++ b/CacheWarmer/HydratorCacheWarmer.php
@@ -50,7 +50,7 @@ class HydratorCacheWarmer implements CacheWarmerInterface
     public function warmUp($cacheDir)
     {
         // we need the directory no matter the hydrator cache generation strategy.
-        $hydratorCacheDir = $this->container->getParameter('doctrine_mongodb.odm.hydrator_dir');
+        $hydratorCacheDir = (string) $this->container->getParameter('doctrine_mongodb.odm.hydrator_dir');
         if (! file_exists($hydratorCacheDir)) {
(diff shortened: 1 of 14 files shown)


CacheWarmer/HydratorCacheWarmer.php
```diff
@@ -23,6 +23,10 @@
  *
  * In the process of generating hydrators the cache for all the metadata is primed also,
  * since this information is necessary to build the hydrators in the first place.
+ *
+ * @internal since version 4.4
+ *
+ * @psalm-suppress ContainerDependency
```
franmomu, line 29: Ref: https://github.com/doctrine/DoctrineMongoDBBundle/pull/683#issuecomment-887085532

I think cache warmer classes are not supposed to be used by end users, this way it would be out of the BC promise and it would be easier to maintain.

`ContainerDependency` comes from `psalm/plugin-symfony` when using the whole container, I've created https://github.com/doctrine/DoctrineMongoDBBundle/issues/688.

Command/LoadDataFixturesDoctrineODMCommand.php
```diff
@@ -87,7 +87,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $dm = $this->getManagerRegistry()->getManager($input->getOption('dm'));
         $ui = new SymfonyStyle($input, $output);
 
-        if ((bool) $input->getOption('services')) {
+        if ($input->getOption('services')) {
```
franmomu, line 90: It was given a RedundantCast error, since it is defined like:

`->addOption('services', null, InputOption::VALUE_NONE, 'Use services as fixtures')`

having `VALUE_NONE` it only can return `boolean`.

PS: Using https://github.com/psalm/psalm-plugin-symfony

Tests/Form/Type/DocumentTypeTest.php
```diff
@@ -112,8 +112,8 @@ public function testManyToManyReferences(): void
         $this->assertInstanceOf(FormView::class, $categoryView);
 
         $this->assertCount(2, $categoryView->children);
-        $this->assertTrue($categoryView->children[0]->vars['checked']);
-        $this->assertFalse($categoryView->children[1]->vars['checked']);
```
franmomu, line 116: `children` should be only accessed by string: https://github.com/symfony/symfony/blob/534e1ed986bc0a33e1093f87bb6cede887a23a79/src/Symfony/Component/Form/FormView.php#L34-L39

I tried `"0"` and `"1"`, but it didn't work I guess it because PHP interprets the numeric integer as integer.

alcaeus (approved):

SenseException (approved):

PR #690, opened 2021-08-06
Basic implementation of psr6 compatibility pass

Ahummeling:
After meeting with @alcaeus we came to the conclusion that it would be best to first implement psr6 compatibility, before moving forward with removing the metadata cache configuration. This pull request is aimed at implementing this compatibility, through means of a compiler pass.

Wanted to get a tiny bit of feedback as I am continuing to figure out which tests cause deprecation triggers, figure out what happens when I install this version in 1 of my projects and all that stuff. 

Will be leaving some comments on my first commit to explain certain choices or things that might appear odd


DependencyInjection/Compiler/CacheCompatibilityPass.php
```diff
@@ -0,0 +1,105 @@
...
+use function assert;
+use function is_a;
+use function is_string;
+use function trigger_deprecation;
+
+/** @internal */
+final class CacheCompatibilityPass implements CompilerPassInterface
+{
+    private const CACHE_SETTER_METHODS_PSR6_SUPPORT = [
+        'setMetadataCache' => true,
```
Ahummeling: Not sure if I should include both of these methods, and also not sure if I have missed any methods. These were the only ones I could think of.

DependencyInjection/Compiler/CacheCompatibilityPass.php
```diff
@@ -0,0 +1,105 @@
...
+        'setMetadataCacheImpl' => true,
+    ];
+
+    public function process(ContainerBuilder $container): void
+    {
+        foreach (array_keys($container->findTaggedServiceIds(DoctrineMongoDBExtension::CONFIGURATION_TAG)) as $id) {
+            /** @var array<int, string|Reference[]> $methodCall */
+            foreach ($container->getDefinition($id)->getMethodCalls() as $methodCall) {
```
Ahummeling, line 31: Really not sure why the dependency-injection component doesn't have a `MethodCall` class.. but basically spent way too long figuring out that `$methodCall` is an array where index 0 points to the method name as a string, and index 1 points to an array of references which represent the arguments of the methodcall. Figured some additional phpdocs wouldn't hurt and neither would the assert call made here, but might be wrong about that :)

DependencyInjection/Compiler/CacheCompatibilityPass.php
```diff
@@ -0,0 +1,105 @@
...
+
+                if (! isset(self::CACHE_SETTER_METHODS_PSR6_SUPPORT[$methodName])) {
+                    continue;
+                }
+
+                /** @var Reference[] $methodArgs */
+                $methodArgs   = $methodCall[1];
+                $definitionId = (string) $methodArgs[0];
+                $aliasId      = null;
+                if ($container->hasAlias($definitionId)) {
```
Ahummeling, line 41: So this is where I deviate the most from the doctrinebundle, because as doctrinebundle expects to get an alias here, I actually didn't find an alias for the method argument. <br>
Figured it wasn't actually relevant for the whole wrapping anyway, so just go with whatever's given and make sure to update the alias, if the argument represents an alias.

DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -37,6 +37,8 @@
  */
 class DoctrineMongoDBExtension extends AbstractDoctrineExtension
 {
+    public const CONFIGURATION_TAG = 'doctrine.odm.configuration';
```
Ahummeling, line 50: doctrine bundle defines this tag in 2 places, as a private const in the compatibilitypass and as a public const in another pass. Imo it makes most sense to define it in 1 place; the extension class, but maybe someone can elaborate on why doctrine bundle implemented it in that way.

Tests/FixtureIntegrationTest.php
```diff
@@ -29,6 +29,7 @@
 use function sprintf;
 use function sys_get_temp_dir;
 
+/** @group legacy */
```
Ahummeling: This is probably not the way to go, but it was a nice way for me to toggle deprecation triggers with phpunit easily. Will figure out the correct approach here

Ahummeling (commented):
When reviewing this PR, please make sure to have a look over my comments as well, as I'm sure they all raise some potential points for (beneficial) discussion / constructive criticism.

franmomu:
Thanks @Ahummeling! In order to fix the build, can you please execute `vendor/bin/phpcbf` to fix the coding standard issues (you can check with `vendor/bin/phpcs` if it is fine).

The failing test should be fixed if you rebase with `4.4.x`.

DependencyInjection/Compiler/CacheCompatibilityPass.php
```diff
@@ -0,0 +1,105 @@
...
+use function is_a;
+use function is_string;
+use function trigger_deprecation;
+
+/** @internal */
+final class CacheCompatibilityPass implements CompilerPassInterface
+{
+    private const CACHE_SETTER_METHODS_PSR6_SUPPORT = [
+        'setMetadataCache' => true,
+        'setMetadataCacheImpl' => true,
```
alcaeus:
```suggestion
```

DependencyInjection/Compiler/CacheCompatibilityPass.php
alcaeus: In this case, we only need to do this for `setMetadataCache`: the extension always configures a call to `setMetadataCache`, regardless of the type of cache involved. The compiler pass then checks the method argument and decides whether the argument needs to be wrapped in an adapter.

If we bump the required version of ODM to 2.3 (which we should do in this PR), we don't have to worry about the method not being there at all, so we can stop using the deprecated API immediately.

DependencyInjection/Compiler/CacheCompatibilityPass.php
alcaeus: You might want to file an issue for this or reach out to some of the core Symfony folks about this. I'm not sure how that could be introduced without BC breaks, but I do agree that using arrays for this stuff seems weird. Maybe @nicolas-grekas can provide some insight into this.

DependencyInjection/Compiler/CacheCompatibilityPass.php
alcaeus: Ping @ostrolucky as he was the one that came up with the logic after I failed to do so. Maybe he can provide some insight into this.

DependencyInjection/Compiler/CacheCompatibilityPass.php
```diff
@@ -0,0 +1,105 @@
...
+        $targetClass   = CacheProvider::class;
+        $targetFactory = DoctrineProvider::class;
+
+        if ($shouldBePsr6) {
+            $targetClass   = CacheItemPoolInterface::class;
+            $targetFactory = CacheAdapter::class;
+
+            trigger_deprecation(
+                'doctrine/doctrine-bundle',
+                '2.4',
```
alcaeus:
```suggestion
                'doctrine/mongodb-odm-bundle',
                '4.4',
```

DependencyInjection/DoctrineMongoDBExtension.php
alcaeus: Agree. Only change I would make is to mark the constant as internal, since we don't consider this part of the public API.
```suggestion
    /** @internal */
    public const CONFIGURATION_TAG = 'doctrine.odm.configuration';
```

Tests/FixtureIntegrationTest.php
alcaeus: Without changes to the configuration, the extension will always call the deprecated method in ODM. This needs changing in `DoctrineMongoDBExtension::loadDocumentManager`, where we add the call to `setMetadataCacheImpl`:
https://github.com/doctrine/DoctrineMongoDBBundle/blob/3f6449eb05ff5395f90d834b3ff87be3b33d62d7/DependencyInjection/DoctrineMongoDBExtension.php#L208

Changing this to `setMetadataCache` (after bumping the ODM dependency to `^2.3`) will fix this.

alcaeus (commented):
Left an initial review, great work so far!

DependencyInjection/Compiler/CacheCompatibilityPass.php
franmomu: In the meantime we can use an array-shape like:

```php
/** @var array{0: string, 1: mixed[]} $methodCall */
```
based on [`addMethodCall`](https://github.com/symfony/symfony/blob/76a414c7d8ad03b5d6e520edfd8b947af21717b0/src/Symfony/Component/DependencyInjection/Definition.php#L353) I think it's not only `Reference[]`, it could be any argument (scalar, arrays, etc).

This way the assert won't be necessary.

I don't know if the Symfony people would accept array-shapes in the phpdoc, that would be quite useful.

DependencyInjection/DoctrineMongoDBExtension.php
Ahummeling: I thought the entire class already marked `@internal`, but I see now that it isn't. Perhaps we should consider marking the class internal instead?

alcaeus: Yes, that makes more sense.

DependencyInjection/Compiler/CacheCompatibilityPass.php
Ahummeling: ++ 
Also bumped version of ODM to 2.3

DependencyInjection/Compiler/CacheCompatibilityPass.php
Ahummeling: Realised that the definitionId should then be the aliasId

Tests/FixtureIntegrationTest.php
Ahummeling: Yeah removed this again, should be correct now.

DependencyInjection/Compiler/CacheCompatibilityPass.php
Ahummeling: ++

DependencyInjection/Compiler/CacheCompatibilityPass.php
Ahummeling: Whoops, fixed that

DependencyInjection/Compiler/CacheCompatibilityPass.php
Ahummeling: That's a great suggestion, and good point. Didn't consider it could be something than `Reference[]`

DependencyInjection/DoctrineMongoDBExtension.php
Ahummeling: After discussion, decided that because bundles are allowed to load other bundles' configuration to allow inter-bundle integration, to make just the const internal.

Ahummeling:
just need to have a testcase added for the compiler pass now and this should be complete

DependencyInjection/Compiler/CacheCompatibilityPass.php
```diff
@@ -0,0 +1,100 @@
...
+        if ($compatibilityLayer === null) {
+            return;
+        }
+
+        $aliasId = $aliasId ?? $definitionId;
+
+        $compatibilityLayerId = $definitionId . '.compatibility_layer';
+        $container->setAlias($compatibilityLayerId, $aliasId);
+
+        $container->setDefinition($compatibilityLayerId, $compatibilityLayer);
```
franmomu: Not sure about this, but if I had understood correctly, if `$compatibilityLayer` is not `null`, it is a definition implementing PSR-6 to replace the current definition, so should we replace the current alias with the compatibilityLayer one (flip arguments)?:

```php
$container->setAlias($aliasId, $compatibilityLayerId);
```

franmomu:
@Ahummeling I've rebased, tried to add tests for the compiler pass (based on the ORM) and fixed some psalm issues, finally I've flipped the alias arguments to make tests pass, but not sure if I was right about that last one 🤔 cc @alcaeus

IonBazan:
I've rebased, resolved conflicts and added a dummy assertion to make PHPUnit report the coverage for that CompilerPass test. 
@franmomu can you help to fix the static analysis issues? Maybe we can just revert these phpdoc changes for now?

franmomu:
Not sure why doesn't infer correctly that class string type 😕, maybe because of generics.

IonBazan:
Thanks, it looks good now!

IonBazan:
Thanks, it looks good now!

@alcaeus this is ready for review now.

IonBazan (approved):

franmomu (approved):

IonBazan:
Thank you @Ahummeling! 🚀

PR #692, opened 2021-08-10
add attribute driver support

IonBazan:
This PR adds support for upcoming attribute mapping driver introduced in https://github.com/doctrine/mongodb-odm/pull/2344.

### DIFF
--- a/Resources/doc/config.rst
+++ b/Resources/doc/config.rst
@@ -126,7 +126,7 @@ Explicit definition of all the mapped documents is the only necessary
 configuration for the ODM and there are several configuration options that you
 can control. The following configuration options exist for a mapping:
 
-- ``type`` One of ``annotation``, ``xml``, ``yml``, ``php`` or ``staticphp``.
+- ``type`` One of ``annotation``, ``attribute``, ``xml``, ``php`` or ``staticphp``.
   This specifies which type of metadata type your mapping uses.
 
 - ``dir`` Path to the mapping or document files (depending on the driver). If
@@ -176,7 +176,7 @@ The following configuration shows a bunch of mapping examples:
                     mappings:
                         MyBundle1: ~
                         MyBundle2: xml
-                        MyBundle3: { type: annotation, dir: Documents/ }
+                        MyBundle3: { type: attribute, dir: Documents/ }
                         MyBundle4: { type: xml, dir: Resources/config/doctrine/mapping }
                         MyBundle5:
                             type: xml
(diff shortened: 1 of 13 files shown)


malarzm:
@IonBazan if you want to see whether static analysis is green or add some test feel free to require ODM's 2.3.x branch in composer.json. Later we'll drop the change once 2.3.0 is tagged properly

malarzm:
Also it'd be great to check and update the docs. They definitely need some love since they still mention `yaml` as one of the mapping options 😅 https://symfony.com/doc/current/bundles/DoctrineMongoDBBundle/config.html#mapping-configuration

malarzm:
For the docs we could also add "PHP (attributes)" section for mapping examples: https://symfony.com/doc/current/bundles/DoctrineMongoDBBundle/first_steps.html#creating-a-document-class

IonBazan:
Alright, I've temporarily added `@dev` flag to the version constraint so it will ignore `minimum-stability` during static analysis. I've also added Attribute mapping examples in several places in docs, mentioning they were added in 4.4. 
I'm thinking whether we can make it more clear how to use the new mapping method 🤔 Maybe put attributes examples as the primary?

franmomu:
As suggested in https://github.com/doctrine/mongodb-odm/pull/2344#discussion_r683974460, it would be nice to add some tests like in https://github.com/doctrine/DoctrineBundle/pull/1322/files

IonBazan:
@franmomu @malarzm I've added some test cases for the new driver configuration ✨ 
Also, rebased and fixed Doctor-RST issues. All green now ✅

Resources/doc/config.rst
```diff
@@ -176,7 +176,7 @@ The following configuration shows a bunch of mapping examples:
                     mappings:
                         MyBundle1: ~
                         MyBundle2: xml
-                        MyBundle3: { type: annotation, dir: Documents/ }
+                        MyBundle3: { type: attribute, dir: Documents/ }
```
franmomu, line 179: I think it would be better if we keep `annotation` example and add a new one for `attribute`

Resources/doc/cookbook/registration_form.rst
```diff
@@ -11,65 +11,122 @@ The User Model
...
+        class User
         {
-            $this->password = sha1($password);
+            /**
+             * @MongoDB\Id
+             */
+            #[MongoDB\Id]
+            protected string $id;
+
+            #[MongoDB\Field(type: 'string')]
```
franmomu, line 95: Maybe in our examples we could suggest to use constants like `Doctrine\ODM\MongoDB\Types\Type::STRING`

franmomu (commented):
LGTM, just some minor comments

Resources/doc/cookbook/registration_form.rst
IonBazan: It should be changed in annotation examples then too but I think this is a candidate for a separate PR.
I don't see the constants usages in ORM docs neither and I don't think it's clear to new developers.

franmomu (approved):
After this one we can maybe think about https://github.com/doctrine/DoctrineMongoDBBundle/pull/703 before any other new PR appears

PR #693, opened 2021-08-10
Deprecate some unused `DoctrineODMCommand` methods

franmomu:
This PR adds some functional tests to commands that are not proxied from `doctrine/mongodb-odm`.

It also removes implementing `ContainerAwareInterface` from `DoctrineODMCommand` and deprecates some methods that are not used anymore.


malarzm (approved):
As in another PR with deprecations: could you please add a `UPGRADE-4.4.md` file mentioning the deprecation and `UPGRADE-5.0.md` mentioning removal? This way the change won't fall through the cracks when we release 5.0 and it's what we try to do with ODM changes.

PR #694, opened 2021-08-11
Enforce typehints in tests

franmomu:
This PR removes a phpcs excluded rule that makes mandatory to have type declarations in tests.

This would be nice to do it in the ODM library as well, but the amount of work is considerable so I guess it would be done by steps.


malarzm (approved):

malarzm:
Thanks @franmomu!

> This would be nice to do it in the ODM library as well, but the amount of work is considerable so I guess it would be done by steps.

Yeah, that'll be a lot of work, since the slack with doing so :)

PR #695, opened 2021-08-11
Deprecate tailable cursor command

franmomu:
Apparently this command is not being used for a while since the code is not compatible with `ext-mongodb`.

As suggested by @jmikola, users can use https://docs.mongodb.com/manual/changeStreams/ instead in newer MongoDB versions.

### DIFF
--- a/Command/TailCursorDoctrineODMCommand.php
+++ b/Command/TailCursorDoctrineODMCommand.php
@@ -19,9 +19,12 @@ use Throwable;
 
 use function sleep;
 use function sprintf;
+use function trigger_deprecation;
 
 /**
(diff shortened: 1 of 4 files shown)


malarzm (approved):
@franmomu could you please add a `UPGRADE-4.4.md` file mentioning the deprecation and `UPGRADE-5.0.md` mentioning removal? This way the change won't fall through the cracks when we release 5.0 and it's what we try to do with ODM changes.

franmomu:
Sure! I didn't see those files and I wasn't sure how the deprecation thing works neither which label use. Thanks!

malarzm (approved):

PR #702, opened 2021-09-17
add Stopwatch support

### LINKED ISSUE #701: Add stopwatch support
This has been implemented in #594 but has been closed. IMHO it should be in a separate command listener and each event should contain `requestId` to support concurrency properly.

IonBazan:
This feature adds Symfony Stopwatch support. Events tracked are grouped into `doctrine_mongodb` category and use `requestId` as unique event name identifier.

image

Fixes #701, based on #594 but extracted this logger into a separate class (SRP).

Breaking changes:
 - `doctrine_mongodb.odm.command_logger` service ID is now `doctrine_mongodb.odm.psr_command_logger`

### DIFF
--- a/DependencyInjection/DoctrineMongoDBExtension.php
+++ b/DependencyInjection/DoctrineMongoDBExtension.php
@@ -14,6 +14,7 @@ use Doctrine\Common\EventSubscriber;
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Symfony\Bridge\Doctrine\DependencyInjection\AbstractDoctrineExtension;
 use Symfony\Bridge\Doctrine\Messenger\DoctrineClearEntityManagerWorkerSubscriber;
+use Symfony\Component\Config\Definition\BaseNode;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Alias;
 use Symfony\Component\DependencyInjection\ChildDefinition;
@@ -29,6 +30,7 @@ use function class_exists;
 use function class_implements;
 use function in_array;
 use function interface_exists;
+use function method_exists;
 use function reset;
 use function sprintf;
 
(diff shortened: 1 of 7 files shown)


Resources/config/mongodb.xml
```diff
@@ -136,7 +136,11 @@
             <argument type="tagged" tag="doctrine_mongodb.odm.command_logger" />
         </service>
 
-        <service id="doctrine_mongodb.odm.command_logger" class="Doctrine\Bundle\MongoDBBundle\APM\PSRCommandLogger" public="false">
+        <service id="doctrine_mongodb.odm.stopwatch_command_logger" class="Doctrine\Bundle\MongoDBBundle\APM\StopwatchCommandLogger" public="false">
+            <argument type="service" id="debug.stopwatch" on-invalid="null" />
+        </service>
+
+        <service id="doctrine_mongodb.odm.psr_command_logger" class="Doctrine\Bundle\MongoDBBundle\APM\PSRCommandLogger" public="false">
```
malarzm, line 145: We can't change service names just like that as it breaks backward compatibility. The service should still be available under it's old name but marked as deprecated, like described in https://symfony.com/blog/new-in-symfony-4-3-deprecating-service-aliases

malarzm (changes requested):

Resources/config/mongodb.xml
IonBazan: Added an alias and marked it as deprecated.

DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -496,4 +507,11 @@ public function getXsdValidationBasePath()
     {
         return __DIR__ . '/../Resources/config/schema';
     }
+
+    private function buildDeprecationArgs(string $version, string $message): array
+    {
+        return method_exists(BaseNode::class, 'getDeprecation')
```
malarzm, line 514: Please leave a `@todo` noting what's the minimal Symfony version we can drop the if, that'll be handy in the future

malarzm (commented):
We need to add a test covering new functionality. Maybe a functional one like https://github.com/doctrine/DoctrineMongoDBBundle/blob/4.4.x/Tests/DataCollector/CommandDataCollectorTest.php ?

DependencyInjection/DoctrineMongoDBExtension.php
IonBazan: Added note about minimal Symfony version supporting this (5.1).

malarzm (approved):

malarzm:
Thanks @IonBazan!

PR #703, opened 2021-09-21
[RFC] change directory structure

IonBazan:
Moving all source files into `src` directory makes it easier to search trough the code since there is no need to exclude vendor dir all the time. Let me know what are your thoughts.

Warning: please note that `Tests` directory has been renamed to lowercase `tests` which might not be supported on some systems. You may need to rename it manually after checking out this branch.


malarzm:
@alcaeus I remember we were talking about this transition at least few times before, were we just procrastinating doing so or was there something different? I recall the argument about rebasing open PRs, but given there are just 4 other so it's not that important :)

franmomu:
I'm 👍, if it's because of the open PRs it not a big deal right now I think

PR #706, opened 2021-10-16
Deprecate configuring metadata cache

franmomu:
Follow-up of https://github.com/doctrine/DoctrineMongoDBBundle/pull/675 and https://github.com/doctrine/DoctrineMongoDBBundle/pull/685

### DIFF
--- a/CacheWarmer/MetadataCacheWarmer.php
+++ b/CacheWarmer/MetadataCacheWarmer.php
@@ -1,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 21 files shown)


franmomu:
On hold because of https://github.com/doctrine/DoctrineBundle/pull/1405

franmomu:
If we do not deprecate it, we still can add the autoconfiguration if the metadata cache is not specified, right? I'll update the PR in the following days

franmomu:
Let's close this for now since it was undeprecated in the ORM.

PR #707, opened 2021-10-16
Remove deprecations using Symfony 5.4

franmomu:
This PR adds a CI job testing Symfony 5.4.

[Symfony 5.4 adds deprecations](https://symfony.com/blog/preparing-your-apps-and-bundles-for-symfony-6#introducing-php-types-everywhere) where there is no return type explicitly set (native or phpdoc) in overriding or implementing methods from Symfony or core.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -26,9 +26,11 @@ jobs:
           - "7.3"
           - "7.4"
           - "8.0"
+        stability:
+          - "stable"
         symfony-version:
           - "4.4.x"
-          - "5.1.x"
+          - "5.3.x"
         driver-version:
           - "stable"
         dependencies:
@@ -38,7 +40,14 @@ jobs:
             os: "ubuntu-20.04"
             php-version: "7.2"
             driver-version: "1.5.0"
+            stability: "stable"
             symfony-version: "4.4.*"
+          - dependencies: "highest"
+            os: "ubuntu-20.04"
+            driver-version: "stable"
+            php-version: "8.0"
+            stability: "dev"
+            symfony-version: "5.4.*"
 
     services:
       mongodb:
@@ -80,7 +89,7 @@ jobs:
         run: "composer global require --no-progress --no-scripts --no-plugins symfony/flex"
 
       - name: "Set minimum-stability to stable in Composer"
-        run: "composer config minimum-stability stable"
+        run: "composer config minimum-stability ${{ matrix.stability }}"
 
       - name: "Install dependencies with Composer"
         uses: "ramsey/composer-install@v1"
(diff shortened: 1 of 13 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -81,6 +90,7 @@ jobs:
 
       - name: "Set minimum-stability to stable in Composer"
         run: "composer config minimum-stability stable"
+        if: "${{ matrix.stability == 'stable' }}"
```
IonBazan:
```suggestion
        run: "composer config minimum-stability ${{ matrix.stability }}"
```

Let's not rely on what's in the `composer.json`. In fact, I'd change that to `stable` there because it was only introduced in to pull ODM v2.3-dev.

Tests/FixtureIntegrationTest.php
```diff
@@ -165,6 +165,10 @@ public function testFixturesLoaderWithGroupsOptionViaTag(): void
             $c->autowire(WithDependenciesFixtures::class)
                 ->addTag(FixturesCompilerPass::FIXTURE_TAG, ['group' => 'group2']);
 
+            $c->autowire(RequiredConstructorArgsFixtures::class)
```
IonBazan: I don't think it actually fixes the issue. For me, it causes some other tests in this class to fail randomly due to the execution order or cache dir conflict.

.github/workflows/continuous-integration.yml
franmomu: 👍 (it was introduced [when testing Symfony 5 by the way](https://github.com/doctrine/DoctrineMongoDBBundle/pull/602))

Tests/FixtureIntegrationTest.php
franmomu: I've changed the approach broadening the randomness of `rand(100, 999)` with `uniqid()`.

IonBazan: I came up with same solution on my local env. I wonder if there is a way to delete the cache if it already exists? Or perhaps we can use `reboot()` method to make sure the cache is rebuilt.

phpunit.xml.dist
```diff
@@ -35,4 +35,8 @@
             </exclude>
         </whitelist>
     </filter>
+
+    <listeners>
+        <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener"/>
```
IonBazan, line 40: What's the reason to add this listener here?

IonBazan (approved):

phpunit.xml.dist
franmomu: To show the deprecation warnings, we are executing `phpunit` (not `simple-phpunit`):

https://github.com/doctrine/DoctrineMongoDBBundle/blob/38695331c7bd01711039b5eb99a929442dafb0f1/.github/workflows/continuous-integration.yml#L90-L96

IonBazan: Nice, let me merge it then.

PR #708, opened 2021-10-29
improve collector view

IonBazan:
This PR adds command time to the toolbar just like in ORM bundle:

image

The `ms` unit has been grayed out in the profiler to match other tabs:
image

I've also skipped check for `profiler_markup_version` as `1` means it's Symfony <2.8 which is no longer supported by the bundle.

I rephrased the "Query count" to "command count" but I'm not really sure how to name the "Query time".

### DIFF
--- a/Resources/views/Collector/mongodb.html.twig
+++ b/Resources/views/Collector/mongodb.html.twig
@@ -2,21 +2,23 @@
 
 {% block toolbar %}
     {% if collector.commandCount > 0 %}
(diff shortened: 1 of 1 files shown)


malarzm:
> I rephrased the "Query count" to "command count" but I'm not really sure how to name the "Query time".

How about "Command time"? :)

malarzm (approved):

franmomu (approved):

PR #710, opened 2021-11-07
Add compatibility with Symfony 6

franmomu:
There are many BC breaks here

- `DoctrineMongoDBExtension`: I think they could be acceptable since I think it's quite unusual that someone extends from it.
- `DoctrineMongoDBBundle`: The same thing, I think it is unlikely that someone extends it.
- `MongoDBQueryBuilderLoader` and `DocumentType`: I'm not sure about these ones, should we comment on https://github.com/symfony/symfony/issues/43021? or do you think it's safe to add the return types.

Tests will be probably fixed with https://github.com/symfony/symfony/pull/43917

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -48,6 +48,12 @@ jobs:
             php-version: "8.0"
             stability: "dev"
             symfony-version: "5.4.*"
+          - dependencies: "highest"
+            os: "ubuntu-20.04"
+            driver-version: "stable"
+            php-version: "8.0"
+            stability: "dev"
+            symfony-version: "6.0.*"
 
     services:
       mongodb:
--- a/DependencyInjection/DoctrineMongoDBExtension.php
+++ b/DependencyInjection/DoctrineMongoDBExtension.php
@@ -459,26 +459,18 @@ class DoctrineMongoDBExtension extends AbstractDoctrineExtension
 
     /**
      * @param string $name
-     *
-     * @return string
      */
-    protected function getObjectManagerElementName($name)
+    protected function getObjectManagerElementName($name): string
     {
         return 'doctrine_mongodb.odm.' . $name;
     }
 
-    /**
-     * @return string
-     */
-    protected function getMappingObjectDefaultName()
+    protected function getMappingObjectDefaultName(): string
     {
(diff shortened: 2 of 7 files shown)


malarzm:
- `DocumentType` should be "extended" in a symfony/form-type-way (`::getParent() method IIRC) so this should be safe.
- `MongoDBQueryBuilderLoader` how this is solved for ORM?

franmomu:
> * `MongoDBQueryBuilderLoader` how this is solved for ORM?

[It's in Symfony](https://github.com/symfony/symfony/blob/6.0/src/Symfony/Bridge/Doctrine/Form/ChoiceList/ORMQueryBuilderLoader.php) 😬

jordisala1991:
@franmomu it does not seem that your PR (https://github.com/symfony/symfony/pull/43917) is getting merged for 5.4 or 6.0 (it is target to 6.1), does that affect this PR in any way?

We only miss this one to add sf 6 to three more Sonata bundles.

franmomu:
> @franmomu it does not seem that your PR ([symfony/symfony#43917](https://github.com/symfony/symfony/pull/43917)) is getting merged for 5.4 or 6.0 (it is target to 6.1), does that affect this PR in any way?

I've written a comment there, thanks for pointing that out. 

In any case, that's when using auto_mapping, we can always set it explicitly, I'll try to change it later today or tomorrow to see if it passes the tests.

malarzm:
@franmomu I think let's add the types everywhere, let's just put it to the UPGRADE file that in an (unlikely ;) ) event somebody did extend our classes, they need to adjust their code. It's a small BC break they'll need to live with :)

franmomu:
> @franmomu I think let's add the types everywhere, let's just put it to the UPGRADE file that in an (unlikely ;) ) event somebody did extend our classes, they need to adjust their code. It's a small BC break they'll need to live with :)

👍 I think it's unlikely to happen too, now we need https://github.com/doctrine/DoctrineMongoDBBundle/pull/713 first

malarzm:
#713 merged :)

franmomu:
All 🟢 now!

malarzm (approved):
🎉

malarzm:
Thanks @franmomu!

PR #711, opened 2021-11-08
Reuse workflows from upstream

greg0ire:


.github/workflows/coding-standards.yml
```diff
@@ -11,31 +11,4 @@ on:
 jobs:
   coding-standards:
     name: "Coding Standards"
-    runs-on: "ubuntu-20.04"
-
-    strategy:
-      matrix:
-        php-version:
-          - "7.4"
-          - "8.0"
```
greg0ire, line 20: I don't think running the CS job with 2 PHP version was actually intended (introduced in #662 )

greg0ire:
What's up with the build?

PR #713, opened 2021-11-13
Add support for detecting bundle mapping with the new bundle structure

mbabker:
Companion PR to https://github.com/symfony/symfony/pull/43181

This adds support for detecting bundle mapping if a bundle is using the newer directory structure.

### DIFF
--- a/Tests/DependencyInjection/AbstractMongoDBExtensionTest.php
+++ b/Tests/DependencyInjection/AbstractMongoDBExtensionTest.php
@@ -23,6 +23,7 @@ use Doctrine\ODM\MongoDB\Mapping\Driver\AttributeDriver;
 use Doctrine\Persistence\Mapping\Driver\MappingDriverChain;
 use MongoDB\Client;
 use PHPUnit\Framework\AssertionFailedError;
+use Symfony\Bridge\Doctrine\SchemaListener\DoctrineDbalCacheAdapterSchemaSubscriber;
 use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntityValidator;
 use Symfony\Component\DependencyInjection\Container;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
@@ -36,6 +37,7 @@ use function array_search;
 use function class_exists;
 use function class_implements;
 use function in_array;
+use function is_dir;
 use function reset;
 
 abstract class AbstractMongoDBExtensionTest extends TestCase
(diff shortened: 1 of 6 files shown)


malarzm:
@mbabker thanks for the PR!

> This adds support for detecting bundle mapping if a bundle is using the newer directory structure.

Could you please link what the newer directory structure looks like? I tried googling but to no avail, from the PR I guess that `Resources` folder is being ditched?

If above is true we have some references in the docs that mapping files should be put in `Resources/config/doctrine/` directory (for instance https://www.doctrine-project.org/projects/doctrine-mongodb-bundle/en/4.3/config.html#mapping-configuration). Could you please write some note that it now varies basing on Symfony's version?

mbabker:
> Could you please link what the newer directory structure looks like? I tried googling but to no avail, from the PR I guess that `Resources` folder is being ditched?

Pretty much the same as a Flex application's structure, see https://symfony.com/doc/current/bundles/best_practices.html#directory-structure

> If above is true we have some references in the docs that mapping files should be put in `Resources/config/doctrine/` directory (for instance https://www.doctrine-project.org/projects/doctrine-mongodb-bundle/en/4.3/config.html#mapping-configuration). Could you please write some note that it now varies basing on Symfony's version?

It's not entirely Symfony version dependent, it's pretty much 100% bundle dependent.  Symfony itself supports the newer root structure since 4.4, but admittedly I haven't seen it used on too many bundles yet (I'm using it on branches I've dropped Symfony 3.4 support because I do think it's a bit cleaner, but that's a subjective opinion at best), and I can't say I've seen anything that implies that the older structure with the `Resources` folder that everyone knows will be going away anytime soon.

The gist of it though is that for bundles, if you use the structure recommended in the best practices guide, when a combination of the Doctrine Bridge PR and the Doctrine bundle PR(s) (both here and the ORM) relevant to your bundle land, and your minimum versions support these PRs, then you can put your mapping files in `config/doctrine` and they'd be automatically picked up the same way that they are from `Resources/config/doctrine` now.  If you're crazy enough to live on the edge and use the newer structure AND want to put the files in `config/doctrine`, you can do it now, but you have to deal with the path registration on your own (see https://github.com/BabDev/MoneyBundle/blob/1.x/src/BabDevMoneyBundle.php for an example).

malarzm:
Thanks for the explanation :) I still think the docs could be updated to reflect that `config/doctrine` will be automatically picked up if possible (and using one or another is up to user's taste)

Tests/DependencyInjection/AbstractMongoDBExtensionTest.php
```diff
@@ -309,6 +310,20 @@ public function testXmlBundleMappingDetection(): void
...
+    {
+        $container = $this->getContainer('NewXmlBundle');
+        $loader    = new DoctrineMongoDBExtension();
+        $config    = DoctrineMongoDBExtensionTest::buildConfiguration(
+            ['document_managers' => ['default' => ['mappings' => ['NewXmlBundle' => []]]]]
+        );
+        $loader->load($config, $container);
+
+        $calls = $container->getDefinition('doctrine_mongodb.odm.default_metadata_driver')->getMethodCalls();
+        $this->assertEquals('doctrine_mongodb.odm.default_xml_metadata_driver', (string) $calls[0][1][0]);
```
franmomu, line 328: I guess this assert is failing because this will only be true when using Symfony >= 5.4 🤔

mbabker: Hmm, I'll have to figure out what's happening differently here compared to DoctrineBundle then because the CI build is green with the same feature PR merged and I didn't add any conditionals to the tests there.

franmomu: hmm comparing with https://github.com/doctrine/DoctrineBundle/pull/1418/files, there is no `testNewBundleStructureXmlBundleMappingDetection` test there

franmomu:
~The failing tests are not related to this PR, I think they will be fixed with https://github.com/doctrine/DoctrineMongoDBBundle/pull/714, can you please add those changes here to have the CI build green? everything else~ LGTM.

franmomu:
Restarting the build

franmomu (approved):

malarzm (approved):

malarzm:
Thanks a lot @mbabker!

PR #715, opened 2021-11-23
add AsDocumentListener attribute

IonBazan:
This PR adds `AsDocumentListener` attribute which adds `doctrine_mongodb.odm.event_listener` tag to the service.

Depends on #716 as using attributes in DI requires class to be found in order to use Reflection.

### DIFF
--- a/Attribute/AsDocumentListener.php
+++ b/Attribute/AsDocumentListener.php
@@ -1,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Bundle\MongoDBBundle\Attribute;
+
+use Attribute;
+
+/**
+ * Service tag to autoconfigure document listeners.
+ */
+#[Attribute(Attribute::TARGET_CLASS | Attribute::IS_REPEATABLE)]
+class AsDocumentListener
+{
+    public function __construct(
+        public ?string $event = null,
+        public ?string $method = null,
+        public ?bool $lazy = null,
+        public ?string $connection = null,
+    ) {
+    }
+}
--- a/DependencyInjection/DoctrineMongoDBExtension.php
+++ b/DependencyInjection/DoctrineMongoDBExtension.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\Bundle\MongoDBBundle\DependencyInjection;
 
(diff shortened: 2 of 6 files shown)


Attribute/AsDocumentListener.php
```diff
@@ -0,0 +1,22 @@
...
+ * Service tag to autoconfigure document listeners.
+ */
+#[Attribute(Attribute::TARGET_CLASS | Attribute::IS_REPEATABLE)]
+class AsDocumentListener
+{
+    public function __construct(
+        public ?string $event = null,
+        public ?string $method = null,
+        public ?bool $lazy = null,
+        public ?string $connection = null,
```
IonBazan, line 19: I'm not really sure if this is correct as this doesn't seem to be documented well. I was expecting passing `document_manager` name instead of connection here.

malarzm:
I'm prolly missing something, but how is this different from https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/2.2/reference/events.html#lifecycle-callbacks?

IonBazan:
@malarzm this is just a way to register an EventListener without touching `services.yaml` - instead of tagging the service there, you can add an attribute to the class. This mirrors the ORM's feature: https://github.com/doctrine/DoctrineBundle/pull/1345 and uses Symfony 5.3 feature: https://symfony.com/blog/new-in-symfony-5-3-service-autoconfiguration-and-attributes

Attribute/AsDocumentListener.php
IonBazan: Checked with https://github.com/symfony/doctrine-bridge/blob/5.3/DependencyInjection/CompilerPass/RegisterEventListenersAndSubscribersPass.php and it seems that it's correct.

malarzm:
Ah cool, somehow I thought I saw in PR `#AsDocumentListener` on a document class and thought about lifecycle callbacks. Must have been late, sorry ^^

Attribute/AsDocumentListener.php
malarzm: Documentation confirms that :) https://www.doctrine-project.org/projects/doctrine-mongodb-bundle/en/4.3/events.html#event-listeners

malarzm (approved):

PR #721, opened 2021-12-03
[doc] Remove Unrecognized option metadata_cache_driver under doctrine_mongodb

sormes:
Version: 4.4.1

https://github.com/doctrine/DoctrineMongoDBBundle/blob/4.4.x/Resources/doc/config.rst

In "Multiple Connections" uses option metadata_cache_driver under doctrine_mongodb but it is not supported,


malarzm:
Right, this should be in a `document_manager` section (as also in the docs). Thanks for fixing this @sormes!

PR #734, opened 2022-05-08
Allow doctrine/persistence 3

franmomu:


malarzm:
I've just released and merged up 4.4.3 so this can be rebased now :)

SenseException:
This probably didn't started with this PR, but the Persistence ^1.3.6 constraint doesn't install in any of the CI builds and isn't tested anymore.

malarzm:
> This probably didn't started with this PR, but the Persistence ^1.3.6 constraint doesn't install in any of the CI builds and isn't tested anymore.

Looks like we've dropped ^1.3.6 in https://github.com/doctrine/mongodb-odm/pull/2304/files which was released as a part of ODM 2.3.0. I guess we should have dropped it with 4.4.0 of the bundle as it started requiring ODM ^2.3.0. @franmomu mind removing it along with this PR?

malarzm:
Actually the ^2.0 part can be bumped as well to match ODM constraint. OTOH we have composer to do the resolving for us so I wouldn't mind keeping stuff as is. Y'all decide :)

franmomu:
sure, I've used `^2.2` as minimum as required in ODM 2.3

malarzm:
Thanks @franmomu!

PR #751, opened 2022-11-30
Update first_steps.rst

bruteforks:
Not sure if this is the right place, but I was just stuck on this issue for a long time before figuring it out.

### DIFF
--- a/Resources/doc/first_steps.rst
+++ b/Resources/doc/first_steps.rst
@@ -46,7 +46,8 @@ For Doctrine to be able to do this, you have to create "metadata", or
 configuration that tells Doctrine exactly how the ``Product`` class and its
 properties should be *mapped* to MongoDB. This metadata can be specified
 in a number of different formats including XML or directly inside the
-``Product`` class via annotations or PHP 8 attributes:
+``Product`` class via annotations or PHP 8 attributes (if using attributes,
+you need to change type: annotation to type: attribute in doctrine_mongodb.yaml): 
 
 .. versionadded:: 4.4
 


SenseException:
Thanks for your PR. Your addition is kind of redundant, because the mentioned types are already named in the [documentation](https://github.com/doctrine/DoctrineMongoDBBundle/blob/4.4.x/Resources/doc/config.rst#mapping-configuration). Maybe it makes sense if you set a link to the mapping part instead.

bruteforks:
> Thanks for your PR. Your addition is kind of redundant, because the mentioned types are already named in the [documentation](https://github.com/doctrine/DoctrineMongoDBBundle/blob/4.4.x/Resources/doc/config.rst#mapping-configuration). Maybe it makes sense if you set a link to the mapping part instead.

Took me a day (although new to symfony and doctrine) to figure out that you have to change that obscure config variable for the package to work out of the box. Hopefully it can save someone the hassle. Found fix on SO, so other people have had the same issue: https://stackoverflow.com/a/74084695/1048291

malarzm:
I think the correct way forward is to make `attribute` the new default (and advertise it in docs, newcomers should use just that), we want to phase out annotations anyway :)

SenseException:
> I think the correct way forward is to make attribute the new default

I agree. I'm not sure if this PR needs to be worked on now, but because a default `attribute` won't happen in 4.4, a link still can help people in the meantime.

franmomu:
I've created https://github.com/symfony/recipes-contrib/pull/1472 so for new installs it would rely on the default mapping type detected in https://github.com/symfony/symfony/blob/27e6eb493ab6324164eb53e71fef4bda7fe798dc/src/Symfony/Bridge/Doctrine/DependencyInjection/AbstractDoctrineExtension.php#L298

franmomu:
I think we can close this one as the initial problem I think it was with the Symfony recipe and it was changed in https://github.com/symfony/recipes-contrib/pull/1472

Docs should be change to promote the use of attributes instead of annotations, so I've created https://github.com/doctrine/DoctrineMongoDBBundle/issues/761

thanks @bruteforks for the PR, feel free to reopen if you want to add the link or open a new one.

bruteforks:
@franmomu no problem. glad it helped!

PR #756, opened 2023-01-06
Bump coding standard to v11

franmomu:
Use the latest `doctrine/coding-standard` version ([it supports `"php": "^7.2 || ^8.0"`](https://github.com/doctrine/coding-standard/blob/eec72ad87d3d5d2e8f2aefb58b3880f019b88a7e/composer.json#L25))

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -11,4 +11,4 @@ on:
 jobs:
   coding-standards:
     name: "Coding Standards"
-    uses: "doctrine/.github/.github/workflows/coding-standards.yml@1.2.0"
+    uses: "doctrine/.github/.github/workflows/coding-standards.yml@3.0.0"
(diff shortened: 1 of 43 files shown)


IonBazan:
@franmomu please rebase - there is a conflict after I merged #755

franmomu:
I have to check why it fails 😕

franmomu:
I guess it's because in:

https://github.com/symfony/symfony/blob/2f62c4a7951ed05e722638beb3dcdea7c650a3a2/src/Symfony/Bridge/Doctrine/DependencyInjection/AbstractDoctrineExtension.php#L320-L326

it doesn't match with `/** @Document **/`, I guess there should be a way to not modify these annotation for now.

IonBazan:
Ouch, yes. I think you can add all test fixtures to ignore the coding standard or at least that particular rule.

IonBazan (approved):
LGTM 👍🏻

IonBazan:
Thanks @franmomu!

PR #757, opened 2023-01-06
Fix some issues in `DoctrineMongoDBTypeGuesser`

franmomu:
When trying psalm 5, it complains about an invalid array offset in `mapping['length']`, so I was going to remove it and found some other things to do in the class.


malarzm (approved):

malarzm:
Thanks @franmomu!

PR #758, opened 2023-01-06
Avoid deprecations in tests using Symfony >= 6.1

franmomu:
Seen in https://github.com/doctrine/DoctrineMongoDBBundle/actions/runs/3854144822/jobs/6567819681#step:10:29

I can extract the second commit in another PR if necessary.


malarzm (approved):

malarzm:
> I can extract the second commit in another PR if necessary.

No need :) Thanks @franmomu!

PR #763, opened 2023-01-13
Drop support of PHP < 7.4

franmomu:
As we did it [in the ODM](https://github.com/doctrine/mongodb-odm/pull/2466) I think we can do the same for the `4.6.0` release.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -22,8 +22,6 @@ jobs:
         os:
           - "ubuntu-20.04"
         php-version:
-          - "7.2"
-          - "7.3"
           - "7.4"
           - "8.0"
           - "8.1"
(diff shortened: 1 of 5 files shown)


SenseException (approved):
I suppose PHP 7 will be completely dropped in 5.0.x.

franmomu:
Looking at the installs https://packagist.org/packages/doctrine/mongodb-odm-bundle/php-stats#4.5, 20% are still using 7.4, so I think we should keep it for now.

IonBazan (approved):
Are there any specific PHP 7.4 features we can use in this bundle after this PR is merged?

franmomu:
> Are there any specific PHP 7.4 features we can use in this bundle after this PR is merged?

Yeah, we can add types to private properties, I'll do it in another PR.

PR #766, opened 2023-02-10
Update composer.json

playmono:
Update `doctrine/mongodb-odm` to 2.4.x in order to match function signatures.

This MR aims to fix this issue
```
Doctrine\ODM\MongoDB\APM\CommandLogger::commandStarted(MongoDB\Driver\Monitoring\CommandStartedEvent $event) should either be compatible with MongoDB\Driver\Monitoring\CommandSubscriber::commandStarted(MongoDB\Driver\Monitoring\CommandStartedEvent $event): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/vendor/doctrine/mongodb-odm/lib/Doctrine/ODM/MongoDB/APM/CommandLogger.php on line 47
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         "php": "^7.2 || ^8.0",
         "ext-mongodb": "^1.5",
         "doctrine/annotations": "^1.13",
-        "doctrine/mongodb-odm": "^2.3",
+        "doctrine/mongodb-odm": "^2.4",
         "doctrine/persistence": "^2.2|^3.0",
         "psr/log": "^1.0|^2.0|^3.0",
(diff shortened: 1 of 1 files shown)


franmomu:
Thanks for the PR @playmono, but this is not necessary. The `"^2.3"` constraint means that the range you  can install is `>=2.3` and `<3.0.0`.

If you're not able to install `doctrine/mongodb-odm` `2.4.3`, you can execute the following command to see why:

```
composer why-not doctrine/mongodb-odm 2.4.3
```

playmono:
@franmomu Yes, you are right! Already fixed in my code but I forgot to close this MR :)

Thank you anyway for helping me :)

PR #768, opened 2023-03-05
Use PHP 7.4 for phpcs

franmomu:
Extracted from https://github.com/doctrine/DoctrineMongoDBBundle/pull/765 so it would be easier to review

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -9,7 +9,7 @@
     <!-- Ignore warnings and show progress of the run -->
     <arg value="np"/>
 
-    <config name="php_version" value="70200"/>
+    <config name="php_version" value="70400"/>
 
     <file>.</file>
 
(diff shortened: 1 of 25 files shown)


SenseException (commented):
PHPCS config still has `<config name="php_version" value="70200"/>`. This PR would be a great opportunity to fix this, but shouldn't the PHP version rather be set in the workflow?

greg0ire:
@SenseException I don't think so: contributors should have the same `php_version` value regarless of what version of PHP they use.

phpcs.xml.dist
```diff
@@ -9,7 +9,7 @@
     <!-- Ignore warnings and show progress of the run -->
     <arg value="np"/>
 
-    <config name="php_version" value="70200"/>
+    <config name="php_version" value="70400"/>
```
franmomu, line 12: thanks for the review @SenseException, the `php_version` is already set here

IonBazan (approved):

SenseException (approved):

PR #773, opened 2023-03-14
Leverage targeted resolvers

### LINKED ISSUE #770: Leverage pinned resolvers
Leverage pinned resolvers as recommended by @nicolas-grekas in https://github.com/doctrine/DoctrineMongoDBBundle/pull/754#issuecomment-1457953551

See https://github.com/symfony/symfony/pull/48992

Feel free to assign yourself to this issue.

franmomu:
Closes https://github.com/doctrine/DoctrineMongoDBBundle/issues/770

Symfony allows array attributes for services tags since 6.2: https://github.com/symfony/symfony/pull/47364

### DIFF
--- a/DependencyInjection/DoctrineMongoDBExtension.php
+++ b/DependencyInjection/DoctrineMongoDBExtension.php
@@ -17,7 +17,6 @@ use Doctrine\Common\EventSubscriber;
 use Doctrine\ODM\MongoDB\DocumentManager;
 use InvalidArgumentException;
 use Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver;
-use Symfony\Bridge\Doctrine\Attribute\MapEntity;
 use Symfony\Bridge\Doctrine\DependencyInjection\AbstractDoctrineExtension;
 use Symfony\Bridge\Doctrine\Messenger\DoctrineClearEntityManagerWorkerSubscriber;
 use Symfony\Component\Cache\Adapter\ApcuAdapter;
@@ -424,17 +423,10 @@ class DoctrineMongoDBExtension extends AbstractDoctrineExtension
             return;
         }
 
-        $container->getDefinition('doctrine_mongodb.odm.entity_value_resolver')->setArgument(2, (new Definition(MapEntity::class))->setArguments([
-            null,
-            null,
-            null,
-            $controllerResolverDefaults['mapping'] ?? null,
-            null,
-            null,
-            null,
-            null,
-            $controllerResolverDefaults['disabled'] ?? false,
-        ]));
+        $mapEntityDefinition = $container->getDefinition('doctrine_mongodb.odm.entity_value_resolver.map_entity');
+
+        $mapEntityDefinition->setArgument('$mapping', $controllerResolverDefaults['mapping'] ?? null);
+        $mapEntityDefinition->setArgument('$disabled', $controllerResolverDefaults['disabled'] ?? false);
     }
 
     /**
--- a/DependencyInjection/DoctrineMongoDBExtension.php
+++ b/DependencyInjection/DoctrineMongoDBExtension.php
@@ -25,6 +25,7 @@ use Symfony\Component\Cache\Adapter\ArrayAdapter;
 use Symfony\Component\Cache\Adapter\MemcachedAdapter;
 use Symfony\Component\Cache\Adapter\RedisAdapter;
 use Symfony\Component\Config\FileLocator;
+use Symfony\Component\Config\Loader\FileLoader;
 use Symfony\Component\DependencyInjection\Alias;
 use Symfony\Component\DependencyInjection\ChildDefinition;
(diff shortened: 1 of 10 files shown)


DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -427,6 +396,47 @@ private function loadMessengerServices(ContainerBuilder $container): void
...
+            null,
+            null,
+            null,
+            $controllerResolverDefaults['mapping'] ?? null,
+            null,
+            null,
+            null,
+            null,
+            $controllerResolverDefaults['disabled'] ?? false,
+        ]));
```
alcaeus: Does the container support named arguments? That might make this a little easier to read.

alcaeus (approved):
LGTM with a minor comment on named args.

Resources/config/value_resolver.xml
```diff
@@ -0,0 +1,18 @@
...
+<container xmlns="http://symfony.com/schema/dic/services"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="doctrine_mongodb.odm.entity_value_resolver" class="Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver">
+            <argument type="service" id="doctrine_mongodb" />
+            <argument type="service" id="doctrine_mongodb.odm.entity_value_resolver.expression_language" on-invalid="ignore" />
+            <tag name="controller.argument_value_resolver" priority="110">
+                <attribute name="name">Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver</attribute>
```
nicolas-grekas, line 13: should be:
`<tag name="Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver" priority="110">controller.argument_value_resolver</tag>`

nicolas-grekas: Actually NO: this name would conflict with https://github.com/doctrine/DoctrineBundle/blob/2.9.x/Resources/config/orm.xml#L194 so we need to find another one

nicolas-grekas: On second though, this name might be the right one. It should allow `#[MapEntity]` to work correctly. I don't know how one would expect this to resolve the conflict when there is both ORM + MongoDB but I'd be tempted to say this should be handled inside `EntityValueResolver` (it could decide on its own which manager to use.)

DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -427,6 +395,40 @@ private function loadMessengerServices(ContainerBuilder $container): void
...
+        }
+
+        if ($controllerResolverDefaults === []) {
+            return;
+        }
+
+        $mapEntityDefinition = $container->getDefinition('doctrine_mongodb.odm.entity_value_resolver.map_entity');
+
+        $mapEntityDefinition->setArgument('$mapping', $controllerResolverDefaults['mapping'] ?? null);
+        $mapEntityDefinition->setArgument('$disabled', $controllerResolverDefaults['disabled'] ?? false);
```
franmomu, line 429: @alcaeus I've registered the service `doctrine_mongodb.odm.entity_value_resolver.map_entity` previously and this is another way to only set some parameters, not sure if this is recommended though since if the name of the argument is changed, this will fail.

Resources/config/value_resolver.xml
franmomu: IIUC using Symfony 6.3 (with pinned resolvers), `#[MapEntity]` extending `ValueResolver` will set `EntityValueResolver::class` as resolver, but there won't be any resolver named `EntityValueResolver::class`, even if you are only using ORM, there will be `doctrine.orm.entity_value_resolver` but not `EntityValueResolver::class`, so using `#[MapEntity]` without specifying the resolver will probably fail?

Update: now I see that the ORM will also set its `doctrine.orm.entity_value_resolver` with name `EntityValueResolver` and there is when the conflict will happen.

nicolas-grekas: The resolver it registered under its FQCN, see name attribute on the tag in the XML I linked above.

franmomu:
We are kind of stuck in here, the problem is that with targeted value resolvers, `MapEntity` is tied (by FQCN) to `EntityValueResolver` so we cannot register another one (https://github.com/symfony/symfony/issues/49905)

In our side, the only thing I can think of is to create a `DocumentValueResolver` class which would decorate `EntityValueResolver` and create a `MapDocument` class which would extend `MapEntity`.

boskos-q:
> In our side, the only thing I can think of is to create a `DocumentValueResolver` class which would decorate `EntityValueResolver` and create a `MapDocument` class which would extend `MapEntity`.

@franmomu imo this would make the most sense to do

PR #774, opened 2023-06-29
Add DocumentValueResolver and MapDocument

### LINKED ISSUE #770: Leverage pinned resolvers

franmomu:
Alternative to https://github.com/doctrine/DoctrineMongoDBBundle/pull/773

[Since we cannot use `EntityValueResolver` and `MapEntity`](https://github.com/doctrine/DoctrineMongoDBBundle/pull/773#issuecomment-1548156064), this is an alternative to make it work.

The problem is that since Symfony 6.3, because targeted resolvers, `MapEntity` cannot be used when using ORM + ODM.

Closes #770 
Closes #773

### DIFF
--- a/ArgumentResolver/DocumentValueResolver.php
+++ b/ArgumentResolver/DocumentValueResolver.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Bundle\MongoDBBundle\ArgumentResolver;
+
+use Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Controller\ValueResolverInterface;
+use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
+
+/** @internal */
+final class DocumentValueResolver implements ValueResolverInterface
+{
+    public function __construct(
+        private EntityValueResolver $entityValueResolver,
+    ) {
+    }
+
+    public function resolve(Request $request, ArgumentMetadata $argument): array
+    {
+        return $this->entityValueResolver->resolve($request, $argument);
+    }
+}
--- a/Resources/config/mongodb.xml
+++ b/Resources/config/mongodb.xml
@@ -208,13 +208,5 @@
         <service id="doctrine_mongodb.odm.symfony.fixtures.loader" class="Doctrine\Bundle\MongoDBBundle\Loader\SymfonyFixturesLoader" public="false">
             <argument type="service" id="service_container" />
         </service>
-
-        <service id="doctrine_mongodb.odm.entity_value_resolver" class="Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver">
-            <argument type="service" id="doctrine_mongodb" />
-            <argument type="service" id="doctrine_mongodb.odm.entity_value_resolver.expression_language" on-invalid="ignore" />
-            <tag name="controller.argument_value_resolver" priority="110" />
-        </service>
-
-        <service id="doctrine_mongodb.odm.entity_value_resolver.expression_language" class="Symfony\Component\ExpressionLanguage\ExpressionLanguage" />
     </services>
 </container>
--- a/Resources/config/value_resolver.xml
+++ b/Resources/config/value_resolver.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="doctrine_mongodb.odm.entity_value_resolver" class="Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver">
+            <argument type="service" id="doctrine_mongodb" />
+            <argument type="service" id="doctrine_mongodb.odm.document_value_resolver.expression_language" on-invalid="ignore" />
+        </service>
+
+        <service id="doctrine_mongodb.odm.document_value_resolver.expression_language" class="Symfony\Component\ExpressionLanguage\ExpressionLanguage" />
+
+        <service id="doctrine_mongodb.odm.document_value_resolver" class="Doctrine\Bundle\MongoDBBundle\ArgumentResolver\DocumentValueResolver">
+            <argument type="service" id="doctrine_mongodb.odm.entity_value_resolver" />
+            <tag name="Doctrine\Bundle\MongoDBBundle\ArgumentResolver\DocumentValueResolver" priority="110">controller.argument_value_resolver</tag>
+        </service>
+    </services>
+</container>
--- a/Resources/doc/first_steps.rst
+++ b/Resources/doc/first_steps.rst
@@ -227,7 +227,7 @@ Once you have your repository, you have access to all sorts of helpful methods:
     // find *all* products
     $products = $repository->findAll();
 
-    // find a group of products based on an arbitrary column value
+    // find a group of products based on an arbitrary field value
     $products = $repository->findBy(['price' => 19.99]);
 
 .. note::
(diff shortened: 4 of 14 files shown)


GromNaN:
What is the actual issue that requires a dedicated `DocumentValueResolver`? In your proposition, this resolver is a proxy to `EntityValueResolver` and does nothing different.

The `#[MapEntity]` works as expected in my demo project on Doctrine ODM.
https://github.com/GromNaN/symfony-demo/blob/60eb116cf2b6a0f9acc0bc6591d70f4a4c9f2c69/src/Controller/BlogController.php#L103

Is it to allow using both ORM and ODM in the same project and avoiding the autowiring alias on `Doctrine\Persistence\ManagerRegistry: '@doctrine_mongodb'` :
https://github.com/GromNaN/symfony-demo/blob/60eb116cf2b6a0f9acc0bc6591d70f4a4c9f2c69/config/services.yaml#L40

franmomu:
> Is it to allow using both ORM and ODM in the same project and avoiding the autowiring alias on `Doctrine\Persistence\ManagerRegistry: '@doctrine_mongodb'` : https://github.com/GromNaN/symfony-demo/blob/60eb116cf2b6a0f9acc0bc6591d70f4a4c9f2c69/config/services.yaml#L40

Exactly, [the problem is having both ORM and ODM](https://github.com/doctrine/DoctrineMongoDBBundle/pull/773#discussion_r1148886080) and since `MapEntity` is tied to `EntityValueResolver` by its FQCN I couldn't come up with something better 😞

GromNaN:
An alternative could be to have a `ManagerRegistry` aggregator embedding both ORM and ODM implementations that can be used whether it's an entity or a document.

All in all, I find your approach interesting because it avoids having to use "entity" word for "document". And it doesn't require a lot of specific code.

Resources/config/mongodb.xml
```diff
@@ -208,13 +208,5 @@
         <service id="doctrine_mongodb.odm.symfony.fixtures.loader" class="Doctrine\Bundle\MongoDBBundle\Loader\SymfonyFixturesLoader" public="false">
             <argument type="service" id="service_container" />
         </service>
-
-        <service id="doctrine_mongodb.odm.entity_value_resolver" class="Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver">
```
GromNaN, line 212: It looks like a breaking change if someone is already using the `MapEntity` attribute with the ODM. Not a big deal but it must be documented in the UPGRADE file.

franmomu: hmmm not sure about the BC break, these services were added in `4.6.x` and never been released, so if someone is using it is because `EntityValueResolver` was registered as a service manually, in that case I guess it will continue working fine.

boskos-q (approved):

composer.json
```diff
@@ -41,7 +41,7 @@
         "symfony/stopwatch": "^5.4|^6.2",
         "symfony/validator": "^5.4|^6.2",
```
boskos-q:
```suggestion
        "symfony/stopwatch": "^5.4|^6.3",
        "symfony/validator": "^5.4|^6.3",
```
I don't want to rush ahead of time, but sf 6.2 is supported till the end of July. Maybe to bump version to `^6.3`, tho this can also be done later on as well ;)

Resources/config/mongodb.xml
malarzm: > and never been released

We're fine changing logic given this argument :)

Resources/doc/first_steps.rst
```diff
@@ -249,6 +249,226 @@ to easily fetch objects based on multiple conditions:
...
+
+    #[Route('/product/{id}')]
+    public function showAction(Product $product): Response
+    {
+        // use the Product!
+        // do something, like pass the $product object into a template
+    }
+
+That's it! The bundle uses the ``{id}`` from the route to query for the ``Product``
+by the ``id`` column. If it's not found, a 404 page is generated.
```
malarzm:
```suggestion
by the ``id`` field. If it's not found, a 404 page is generated.
```

Resources/doc/first_steps.rst
```diff
@@ -249,6 +249,226 @@ to easily fetch objects based on multiple conditions:
...
+            User $user
+        ): Response {
+            // User is not resolved by the DocumentValueResolver
+            // ...
+        }
+
+Fetch Automatically
+~~~~~~~~~~~~~~~~~~~
+
+If your route wildcards match properties on your document, then the resolver
```
malarzm:
```suggestion
If your route wildcards match properties in your document, then the resolver
```

not 100% about this one, but `on` sounds weird when translating in my head :)

Resources/doc/first_steps.rst
```diff
@@ -249,6 +249,226 @@ to easily fetch objects based on multiple conditions:
...
+will automatically fetch them:
+
+.. code-block:: php
+
+    /**
+     * Fetch via primary key because {id} is in the route.
+     */
+    #[Route('/product/{id}')]
+    public function showByPk(Post $post): Response
```
malarzm: MongoDB does not have a `primary key` concept, should we just say `identifier` or is there more to itfunctionality-wise?

Resources/doc/first_steps.rst
```diff
@@ -249,6 +249,226 @@ to easily fetch objects based on multiple conditions:
...
+
+MapDocument Options
+~~~~~~~~~~~~~~~~~~~
+
+A number of options are available on the ``MapDocument`` attribute to
+control behavior:
+
+``id``
+    If an ``id`` option is configured and matches a route parameter, then
+    the resolver will find by the primary key:
```
malarzm:
```suggestion
    the resolver will find by the identifier:
```

I've seen `primary key` in previous chapter as well, best to check for it globally

Resources/doc/first_steps.rst
```diff
@@ -249,6 +249,226 @@ to easily fetch objects based on multiple conditions:
...
+            \DateTime $date
+        ): Response {
+        }
+
+``stripNull``
+    If true, then when ``findOneBy()`` is used, any values that are
+    ``null`` will not be used for the query.
+
+``objectManager``
+    By default, the ``DocumentValueResolver`` uses the *default*
```
malarzm: Not sure how this works in ORM but I'd expect passing entity/document to a `ManagerRegistry` to figure out which object manager should be used by default?

Resources/doc/first_steps.rst
```diff
@@ -249,6 +249,226 @@ to easily fetch objects based on multiple conditions:
...
+        #[Route('/product/{id}')]
+        public function show(
+            #[MapDocument(objectManager: 'foo')]
+            Product $product
+        ): Response {
+        }
+
+``evictCache``
+    If true, forces Doctrine to always fetch the document from the database
+    instead of cache.
```
malarzm: ODM does not provide native cache solutions, I believe we shouldn't be documenting (or exposing) the functionality for now

malarzm (commented):
Would it be possible to have few functional tests here in the bundle? Just so we know that decorating original classes still works

Resources/doc/first_steps.rst
franmomu: Yes, you're right, I've removed the documentation about this and cannot be configured

Resources/doc/first_steps.rst
franmomu: https://github.com/symfony/symfony/blob/503a7b3cb62fb6de70176b07bd1c4242e3addc5b/src/Symfony/Bridge/Doctrine/ArgumentResolver/EntityValueResolver.php#L82-L95 I think you mean this, if no parameter is passed, it does what you said.

malarzm: Cool, so the feature works as I've expected it's just the wording that confused me as `default` is often name of a default OM. How about

```suggestion
    By default, the ``DocumentValueResolver`` will choose DocumentManager that has the class registered for it
```

or maybe we have something better for when we're talking about multiple document managers? We rarely do (or at least did when I was working in such environment) so I'm not sure

Resources/config/value_resolver.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="doctrine_mongodb.odm.entity_value_resolver" class="Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver">
```
malarzm, line 8: Should this and other services use `document` instead of `entity`?

Resources/doc/first_steps.rst
malarzm: The method name still uses `Pk` part :)

Tests/DocumentValueResolverFunctionalTest.php
```diff
@@ -0,0 +1,150 @@
...
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\HttpKernel\Kernel;
+use Symfony\Component\Routing\Loader\Configurator\RoutingConfigurator;
+
+use function sys_get_temp_dir;
+use function uniqid;
+
+/**
+ * @requires PHP 8.0
+ * @requires function \Symfony\Bridge\Doctrine\Attribute\MapEntity::__construct
```
malarzm, line 25: oh this is cool, didn't know one can use `@requires` this way!

malarzm (approved):
Thanks for introducing full-blown functional tests! 🎉

Resources/doc/first_steps.rst
franmomu: Oops, thanks! fixed

Resources/doc/first_steps.rst
franmomu: Changed to what you suggested, I haven't found any text in the docs, I think it is fine now and more understandable, thanks

Resources/config/value_resolver.xml
franmomu: I've changed the other ones, but not sure about this one since it's for the `EntityValueResolver`

Tests/DocumentValueResolverFunctionalTest.php
franmomu: https://docs.phpunit.de/en/9.6/incomplete-and-skipped-tests.html#incomplete-and-skipped-tests-requires-tables-api yeah, never used before to be honest 😬

composer.json
franmomu: thanks for the suggestion! but yeah, better to do that in another PR later

ArgumentResolver/DocumentValueResolver.php
```diff
@@ -0,0 +1,24 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\Bundle\MongoDBBundle\ArgumentResolver;
+
+use Symfony\Bridge\Doctrine\ArgumentResolver\EntityValueResolver;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Controller\ValueResolverInterface;
+use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
+
+/** @internal */
```
pepeh, line 12: Why It's marked as `@internal` if the original EntityValueResolver  is not `@internal`?

franmomu: Hi @pepeh, in this case we provide the feature of fetching objects from the controller automatically, I decided to mark this class as `@internal` since to me it was an implementation detail, if we could use the `EntityValueResolver`, we could remove this class.

Are you using this class directly?

pepeh: @franmomu there is option to specify value resolver manually 
`#[ValueResolver(DocumentValueResolver::class)]`
https://symfony.com/doc/current/controller/value_resolver.html#managing-value-resolvers
But if class is `@internal` IDE will show warning

PR #775, opened 2023-07-20
Fix missing driver instrumentation

alcaeus:
The ODM sends its own version data to servers to augment the driver version info. When using the bundle, the client isn't created by the DocumentManager, so this driver data is missing. This PR fixes this issue by adding the correct version info for Doctrine ODM.

Note that the added dependency is already a dependency of the ODM itself and does not add to the dependency chain. It is merely added here to avoid relying on a transitive dependency.

### DIFF
--- a/DependencyInjection/DoctrineMongoDBExtension.php
+++ b/DependencyInjection/DoctrineMongoDBExtension.php
@@ -16,6 +16,7 @@ use Doctrine\Common\DataFixtures\Loader as DataFixturesLoader;
 use Doctrine\Common\EventSubscriber;
 use Doctrine\ODM\MongoDB\DocumentManager;
 use InvalidArgumentException;
+use Jean85\PrettyVersions;
 use Symfony\Bridge\Doctrine\DependencyInjection\AbstractDoctrineExtension;
 use Symfony\Bridge\Doctrine\Messenger\DoctrineClearEntityManagerWorkerSubscriber;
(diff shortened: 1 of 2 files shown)


alcaeus:
Note that a number of check failures are unrelated to the current. I'd like to revisit this in the 4.6.x branch.

DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -407,6 +412,11 @@ private function normalizeDriverOptions(array $connection)
             $driverOptions['context'] = new Reference($driverOptions['context']);
         }
 
+        $driverOptions['driver'] = [
+            'name' => 'symfony-mongodb',
+            'version' => self::getODMVersion(),
```
malarzm, line 417: would it have more value if augmented with bundle's version as well?

malarzm (approved):

DependencyInjection/DoctrineMongoDBExtension.php
alcaeus: I don't think so. I'm not sure what other hosters use this for, but at MongoDB we use the data from MongoDB Atlas to judge version adoption and to influence strategic decisions (e.g. around what PHP versions to support/not support). In that sense the relevant info for us is a) the fact that someone is using the ODM through Symfony, and b) what ODM version they are using.

PR #777, opened 2023-07-26
fix: UniqueEntity::$service is considered final

### LINKED ISSUE #741: UniqueEntity::$service is considered final
Running a test suite in a Symfony 6.1 application throws a deprecation notice:
```
The "Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity::$service" property is considered final. You should not override it in "Doctrine\Bundle\MongoDBBundle\Validator\Constraints\Unique".
```

This is because `Doctrine\Bundle\MongoDBBundle\Validator\Constraints\Unique` class overrides `$service` property with a different default value.

Another way to override it would be to copy the constructor parameters and set the default value there.

https://github.com/doctrine/DoctrineMongoDBBundle/blob/4.6.x/Validator/Constraints/Unique.php

divine:
Fix #741.

### DIFF
--- a/Validator/Constraints/Unique.php
+++ b/Validator/Constraints/Unique.php
@@ -17,19 +17,20 @@ use Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity;
 class Unique extends UniqueEntity
 {
     /**
-     * @param array|string      $fields     The combination of fields that must contain unique values or a set of options
-     * @param bool|array|string $ignoreNull The combination of fields that ignore null values
+     * @param string[]|string      $fields     The combination of fields that must contain unique values or a set of options
+     * @param bool|string[]|string $ignoreNull The combination of fields that ignore null values
+     * @param mixed                $payload
      */
     public function __construct(
         $fields,
-        string $message = null,
+        ?string $message = null,
         string $service = 'doctrine_odm.mongodb.unique',
-        string $em = null,
-        string $entityClass = null,
-        string $repositoryMethod = null,
-        string $errorPath = null,
-        bool|string|array $ignoreNull = null,
-        array $groups = null,
+        ?string $em = null,
+        ?string $entityClass = null,
+        ?string $repositoryMethod = null,
+        ?string $errorPath = null,
+        $ignoreNull = null,
+        ?array $groups = null,
         $payload = null,
         array $options = []
     ) {
@@ -44,7 +45,7 @@ class Unique extends UniqueEntity
             $ignoreNull,
             $groups,
             $payload,
-            $options
+            $options,
         );
     }
 }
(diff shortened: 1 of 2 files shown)


franmomu:
Thanks @divine! what about using the suggested approach by [symfony](https://github.com/symfony/symfony/blob/73e65dc8e745b4f696db4a9631c4d866ab248a8e/UPGRADE-6.1.md#all-components): 

> Public and protected properties are now considered final; instead of overriding a property, consider setting its value in the constructor

That way a user can use `Unique` with their own service without overriding the class (I guess it unlikely to happen, but it could happen)

divine:
Does it look better now?

Thanks!

Validator/Constraints/Unique.php
```diff
@@ -16,6 +16,35 @@
...
+        string $errorPath = null,
+        bool|string|array $ignoreNull = null,
+        array $groups = null,
+        $payload = null,
+        array $options = []
+    ) {
+        parent::__construct(
+            $fields,
+            $message,
+            'doctrine_odm.mongodb.unique',
```
franmomu:
```suggestion
        string $service = 'doctrine_odm.mongodb.unique',
        string $em = null,
        string $entityClass = null,
        string $repositoryMethod = null,
        string $errorPath = null,
        bool|string|array $ignoreNull = null,
        array $groups = null,
        $payload = null,
        array $options = []
    ) {
        parent::__construct(
            $fields,
            $message,
            $service,
```

This way the user can specify the service, otherwise the one from the constructor is not used

Validator/Constraints/Unique.php
```diff
@@ -16,6 +16,35 @@
...
+     */
+    public function __construct(
+        $fields,
+        string $message = null,
+        string $service = 'doctrine_odm.mongodb.unique',
+        string $em = null,
+        string $entityClass = null,
+        string $repositoryMethod = null,
+        string $errorPath = null,
+        bool|string|array $ignoreNull = null,
```
franmomu:
```suggestion
     */
    public function __construct(
        $fields,
        string $message = null,
        string $service = 'doctrine_odm.mongodb.unique',
        string $em = null,
        string $entityClass = null,
        string $repositoryMethod = null,
        string $errorPath = null,
        bool $ignoreNull = null,
```
Now I see this, union types are introduced in PHP >= 8.0, so I guess we should stick to Symfony 5.4 constructor: https://github.com/symfony/symfony/blob/8492f10d29e929a83ccf6c306d40dbad69c9c80b/src/Symfony/Bridge/Doctrine/Validator/Constraints/UniqueEntity.php#L55

~Adding a "simple" test would be awesome, we would have realized of this.~ (Test added)

GromNaN (approved):

malarzm (approved):

malarzm:
Thanks @divine and @franmomu!

divine:
Hello,

Any ETA for a new release with this fix?

Thanks!

malarzm:
Here you go :) https://github.com/doctrine/DoctrineMongoDBBundle/releases/tag/4.6.1

PR #778, opened 2023-07-29
Prepare tests for PHPUnit 10

franmomu:
This is something we have to do even if [Symfony PHPUnit bridge does not work yet with PHPUnit 10](https://github.com/symfony/symfony/issues/49069)


IonBazan (approved):

alcaeus (approved):
LGTM!

PR #779, opened 2023-07-31
Fix phpdoc return type

franmomu:
spotted in https://github.com/doctrine/DoctrineMongoDBBundle/actions/runs/5711475720/job/15473160391

https://github.com/symfony/symfony/blob/3034e8e423d0b160c2b44a170016377bd3a59ec7/src/Symfony/Component/DependencyInjection/ContainerAwareTrait.php#L19-L30

`$container` can be null


malarzm (approved):

malarzm:
Thanks @franmomu!

PR #784, opened 2023-10-25
Symfony 7 support

kbond:
Blockers:
- [x] https://github.com/doctrine/mongodb-odm/pull/2558
- [x] https://github.com/vimeo/psalm/pull/10335
- [x] https://github.com/psalm/psalm-plugin-symfony/pull/323

PR #789, opened 2023-10-29
Remove deprecations using Symfony 6.4

franmomu:
Because of https://github.com/symfony/symfony/pull/50391 there are these deprecations:

```
1x: The "Doctrine\Bundle\MongoDBBundle\CacheWarmer\HydratorCacheWarmer::warmUp()" method will require a new "string|null $buildDir" argument in the next major version of its interface "Symfony\Component\HttpKernel\CacheWarmer\WarmableInterface", not defining it is deprecated.
    1x in HydratorCacheWarmerTest::setUp from Doctrine\Bundle\MongoDBBundle\Tests\CacheWarmer

  1x: The "Doctrine\Bundle\MongoDBBundle\CacheWarmer\PersistentCollectionCacheWarmer::warmUp()" method will require a new "string|null $buildDir" argument in the next major version of its interface "Symfony\Component\HttpKernel\CacheWarmer\WarmableInterface", not defining it is deprecated.
    1x in PersistentCollectionCacheWarmerTest::setUp from Doctrine\Bundle\MongoDBBundle\Tests\CacheWarmer

  1x: The "Doctrine\Bundle\MongoDBBundle\CacheWarmer\ProxyCacheWarmer::warmUp()" method will require a new "string|null $buildDir" argument in the next major version of its interface "Symfony\Component\HttpKernel\CacheWarmer\WarmableInterface", not defining it is deprecated.
    1x in ProxyCacheWarmerTest::setUp from Doctrine\Bundle\MongoDBBundle\Tests\CacheWarmer
```

I've created a second commit to silence these ones, but the changes are not BC, so we have to decide if we are fine with that or we should ignore it for now.

There are some other deprecations that we have to take a look.

### DIFF
--- a/Tests/Fixtures/Cache/Collections.php
+++ b/Tests/Fixtures/Cache/Collections.php
@@ -8,23 +8,18 @@ use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;
 use MongoDB\BSON\ObjectId;
 
-/** @ODM\Document */
 #[ODM\Document]
 class Collections
 {
-    /** @ODM\Id */
     #[ODM\Id]
     public ?ObjectId $id = null;
 
-    /** @ODM\EmbedMany(collectionClass=SomeCollection::class) */
     #[ODM\EmbedMany(collectionClass: SomeCollection::class)]
     public SomeCollection $coll;
 
-    /** @ODM\ReferenceMany(collectionClass=SomeCollection::class) */
     #[ODM\ReferenceMany(collectionClass: SomeCollection::class)]
     public SomeCollection $refs;
 
-    /** @ODM\EmbedMany(collectionClass=AnotherCollection::class) */
     #[ODM\EmbedMany(collectionClass: AnotherCollection::class)]
     public AnotherCollection $another;
 }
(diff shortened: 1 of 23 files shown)


Tests/Fixtures/Cache/Collections.php
```diff
@@ -9,18 +9,23 @@
 use MongoDB\BSON\ObjectId;
 
 /** @ODM\Document */
```
GromNaN: Since the TestKernel configuration changed to `attribute`, is it necessary to keep legacy `annotations`?

franmomu: Nope, I was wondering if we should still test `annotations` when using Symfony 5.4 and then remove them when deprecating our `doctrine_mongodb.odm.metadata.annotation*` services

franmomu:
The BC break changes are mandatory if we want to have compatibility with Symfony 7, that's why I targeted `4.7.x`, this PR would resolve some of the issues in https://github.com/doctrine/DoctrineMongoDBBundle/pull/784

Tests/Fixtures/Cache/Collections.php
franmomu: I've added another commit removing the annotations 👍

GromNaN (approved):

GromNaN:
Thank you very much @franmomu

PR #790, opened 2023-11-02
Deprecate `fixture_loader` option

franmomu:
[`ContainerAwareLoader` class is deprecated in Symfony 6.4](https://github.com/symfony/symfony/blob/8a69f6766815bd457ce12ab6a04293207efcec52/src/Symfony/Bridge/Doctrine/DataFixtures/ContainerAwareLoader.php#L19) (and removed in 7.0) and looking at the usages I bumped into `fixture_loader` option.

This option was introduced long time ago in https://github.com/doctrine/DoctrineMongoDBBundle/commit/5a1208c18f77fa9d85420d68a3ad40533133f5ce and it was used [to instantiate that fixture loader in `LoadDataFixturesDoctrineODMCommand`](https://github.com/doctrine/DoctrineMongoDBBundle/commit/5a1208c18f77fa9d85420d68a3ad40533133f5ce#diff-3a492b7fded88ef81783ee4a2b12b644854d41a3caf1944dbafb84f8b7dd594fR100).

Right now this option apparently is only used to set this parameter:
https://github.com/doctrine/DoctrineMongoDBBundle/blob/e8bc04817874b9b63971fd3cea5f1d1dac145b30/DependencyInjection/DoctrineMongoDBExtension.php#L98-L99

But we don't use this parameter since 3.6 in https://github.com/doctrine/DoctrineMongoDBBundle/pull/529 (correction: in 3.6 was still used but in a deprecated path that was removed in 4.0)

I decided to apart from deprecating the option, to remove the parameter (I can re-add it) which I guess/hope no one is using it (they shouldn't).


malarzm (approved):
> I guess/hope no one is using it (they shouldn't).

Let's add our reasoning to UPDATE-4.7md file and I'm willing to take a chance on removing this now 👍

malarzm (approved):

malarzm:
Thanks @franmomu!

PR #793, opened 2023-12-09
Remove support of annotations

franmomu:

PR #794, opened 2023-12-09
Remove unused mapping services

franmomu:
Follow-up of https://github.com/doctrine/DoctrineMongoDBBundle/pull/793

While removing annotation mapping support in https://github.com/doctrine/DoctrineMongoDBBundle/pull/793, I've notice that these services are not necessary.

So mapping drivers are registered in [`AbstractDoctrineExtension::registerMappingDrivers()`](https://github.com/symfony/symfony/blob/fec89090f178f32b686099bec5b2d1521673c514/src/Symfony/Bridge/Doctrine/DependencyInjection/AbstractDoctrineExtension.php#L184).

For example for `attribute` it executes [this code](https://github.com/symfony/symfony/blob/fec89090f178f32b686099bec5b2d1521673c514/src/Symfony/Bridge/Doctrine/DependencyInjection/AbstractDoctrineExtension.php#L205-L207) which in our case uses `doctrine_mongodb.odm.metadata.attribute.class` to create the AttributeDriver service.

### DIFF
--- a/Resources/config/mongodb.xml
+++ b/Resources/config/mongodb.xml
@@ -32,7 +32,6 @@
         <!-- metadata -->
         <parameter key="doctrine_mongodb.odm.metadata.driver_chain.class">Doctrine\Persistence\Mapping\Driver\MappingDriverChain</parameter>
         <parameter key="doctrine_mongodb.odm.metadata.attribute.class">Doctrine\ODM\MongoDB\Mapping\Driver\AttributeDriver</parameter>
-        <parameter key="doctrine_mongodb.odm.metadata.attribute_reader.class">Doctrine\ODM\MongoDB\Mapping\Driver\AttributeReader</parameter>
         <parameter key="doctrine_mongodb.odm.metadata.xml.class">Doctrine\Bundle\MongoDBBundle\Mapping\Driver\XmlDriver</parameter>
 
         <!-- directories -->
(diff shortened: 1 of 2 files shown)


Resources/config/mongodb.xml
```diff
@@ -32,7 +32,6 @@
         <!-- metadata -->
         <parameter key="doctrine_mongodb.odm.metadata.driver_chain.class">Doctrine\Persistence\Mapping\Driver\MappingDriverChain</parameter>
         <parameter key="doctrine_mongodb.odm.metadata.attribute.class">Doctrine\ODM\MongoDB\Mapping\Driver\AttributeDriver</parameter>
```
GromNaN, line 34: This 2 parameters are no longer necessary. 
```suggestion
```

You should also remove the assertions on this parameters in the tests.

GromNaN (changes requested):

Resources/config/mongodb.xml
franmomu: They are still necessary, the thing about how it works is that (for `$driverType = 'attribute'` for example) [this code](https://github.com/symfony/symfony/blob/fec89090f178f32b686099bec5b2d1521673c514/src/Symfony/Bridge/Doctrine/DependencyInjection/AbstractDoctrineExtension.php#L205-L207):

```php
$mappingDriverDef = new Definition($this->getMetadataDriverClass($driverType), [
    array_values($driverPaths),
]);
```
calls to [our](https://github.com/doctrine/DoctrineMongoDBBundle/blob/3c499fb71ab4df6e20cce271e86d3670631c6fd4/DependencyInjection/DoctrineMongoDBExtension.php#L547-L550):
```php
protected function getMetadataDriverClass(string $driverType): string
{
    return '%' . $this->getObjectManagerElementName('metadata.' . $driverType . '.class') . '%';
}

// ...

protected function getObjectManagerElementName(string $name): string
{
    return 'doctrine_mongodb.odm.' . $name;
}
```
so it looks for `%doctrine_mongodb.odm.metadata.attribute.class%`

GromNaN (approved):

Resources/config/mongodb.xml
GromNaN: Ok, got it.

PR #796, opened 2023-12-09
Mark all command and compiler pass classes as final

GromNaN:
All the commands are glue code for Doctrine features. If some needs to change the behavior of a command, they can duplicate the code.
Maintenance is a lot easier when this classes are marked internal, as we don't have to keep backward-compatibility.

### DIFF
--- a/Command/ClearMetadataCacheDoctrineODMCommand.php
+++ b/Command/ClearMetadataCacheDoctrineODMCommand.php
@@ -14,7 +14,7 @@ use Symfony\Component\Console\Output\OutputInterface;
  *
  * @internal since version 4.7.0
  */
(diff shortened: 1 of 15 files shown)


franmomu (approved):
I'm going to open an issue in `mongodb-odm` to add the `dm` option to these commands, that way I think we can remove these classes.

GromNaN:
@franmomu maybe we should mark them internal in 4.7 and final in 5.0. So this breaking change can be reported before upgrading to the major version.

franmomu:
> @franmomu maybe we should mark them internal in 4.7 and final in 5.0. So this breaking change can be reported before upgrading to the major version.

yeah, we should warn users 👍

GromNaN:
The classes are now internal in 4.7, and this PR mark them as final in 5.0

PR #797, opened 2023-12-09
Add return types required by extending Symfony classes

GromNaN:
Cherry-picked from #784 and updated after review.

This types are inherited from parent class in Symfony.

### DIFF
--- a/Command/DoctrineODMCommand.php
+++ b/Command/DoctrineODMCommand.php
@@ -22,19 +22,14 @@ abstract class DoctrineODMCommand extends Command
     }
 
     /** @param string $dmName */
-    public static function setApplicationDocumentManager(Application $application, $dmName)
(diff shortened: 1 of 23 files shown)


Command/DoctrineODMCommand.php
```diff
@@ -22,19 +22,14 @@ public function __construct(private ManagerRegistry $registry)
     }
 
     /** @param string $dmName */
-    public static function setApplicationDocumentManager(Application $application, $dmName)
+    public static function setApplicationDocumentManager(Application $application, $dmName): void
```
franmomu, line 25:
```suggestion
    public static function setApplicationDocumentManager(Application $application, string $dmName): void
```

I'll check these native types in another PR as I guess we have some more we can add, so feel free to ignore this comment in this PR.

franmomu (approved):

PR #798, opened 2023-12-09
Revert "Add $buildDir parameter to warmUp method"

franmomu:
This reverts commit dc54819096a2e57b4d04dca3fd9e275d37024e46.

Since we are not adding BC breaks in `4.7`, should we maybe revert this commit? which was the only one adding a BC break in https://github.com/doctrine/DoctrineMongoDBBundle/pull/789

### DIFF
--- a/CacheWarmer/HydratorCacheWarmer.php
+++ b/CacheWarmer/HydratorCacheWarmer.php
@@ -46,7 +46,7 @@ class HydratorCacheWarmer implements CacheWarmerInterface
     }
 
     /** @return string[] */
-    public function warmUp(string $cacheDir, ?string $buildDir = null)
+    public function warmUp(string $cacheDir)
     {
         // we need the directory no matter the hydrator cache generation strategy.
         $hydratorCacheDir = (string) $this->container->getParameter('doctrine_mongodb.odm.hydrator_dir');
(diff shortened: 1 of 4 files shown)


GromNaN:
Sure. And let's re-apply in 5.0.x

GromNaN (approved):

GromNaN:
Actually, cache warmers should be `@internal` too.

franmomu:
> Actually, cache warmers should be `@internal` too.

Yep, I'm make them and ignore the direct deprecation

franmomu:
> > Actually, cache warmers should be `@internal` too.
> 
> Yep, I'm make them and ignore the direct deprecation

They already are

GromNaN:
> They already are

Then, you don't have to revert. Changing internal classes is never a breaking change.

franmomu:
> > They already are
> 
> Then, you don't have to revert. Changing internal classes is never a breaking change.

But it was marked as internal in `4.4`, someone could upgrade from `4.3` to `4.7`

GromNaN:
Alright.

GromNaN:
Thank you @franmomu

GromNaN:
There was a conflict when upmerging into 5.0.x, I kept the modified changes, so we don't have to reapply this breaking changes for the new major version. See https://github.com/doctrine/DoctrineMongoDBBundle/commit/a4903da062b1f23b8672a87afc793aa19c5bd9b0

PR #799, opened 2023-12-09
Remove `--service` option from `doctrine:mongodb:fixtures:load` command

### LINKED ISSUE #650: Drop --services option for command loading fixtures
It seems like this was supposed to happen in 4.x but fell through the cracks. Cross-linking #649 that made ma realize this

GromNaN:
Fix #650
This option was deprecated, using services is the only behavior, used by default.

### DIFF
--- a/Command/LoadDataFixturesDoctrineODMCommand.php
+++ b/Command/LoadDataFixturesDoctrineODMCommand.php
@@ -7,20 +7,15 @@ namespace Doctrine\Bundle\MongoDBBundle\Command;
 use Doctrine\Bundle\MongoDBBundle\Loader\SymfonyFixturesLoaderInterface;
 use Doctrine\Bundle\MongoDBBundle\ManagerRegistry;
 use Doctrine\Common\DataFixtures\Executor\MongoDBExecutor;
-use Doctrine\Common\DataFixtures\Loader;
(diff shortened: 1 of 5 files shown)


Command/LoadDataFixturesDoctrineODMCommand.php
```diff
@@ -7,42 +7,32 @@
...
 /**
  * Load data fixtures from bundles.
  */
 class LoadDataFixturesDoctrineODMCommand extends DoctrineODMCommand
 {
-    public function __construct(?ManagerRegistry $registry = null, ?KernelInterface $kernel = null, private ?SymfonyFixturesLoaderInterface $fixturesLoader = null)
+    public function __construct(ManagerRegistry $registry, private SymfonyFixturesLoaderInterface $fixturesLoader)
     {
         parent::__construct($registry);
     }
```
GromNaN, line 31: The command is removed in the DI extension instead of being disabled, so the code is more centralized and the dependency to the loader can be required in the constructor.

franmomu:
LGTM after removing the code in https://github.com/doctrine/DoctrineMongoDBBundle/pull/799/files#diff-3a492b7fded88ef81783ee4a2b12b644854d41a3caf1944dbafb84f8b7dd594fR69 (the one that psalm is complaining)

PR #801, opened 2023-12-10
Deprecate container injection in fixture classes with `ContainerAwareInterface`

GromNaN:
[`ContainerAwareInterface`](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/DependencyInjection/ContainerAwareInterface.php) is deprecated since Symfony 6.4. It was used before, for fixture classes that were not defined as services. As all fixture classes must declared as services to be used by the command, proper dependency injection is easy to do.

Let's deprecate this automatic injection and remove it in 5.0 #802

### DIFF
--- a/Loader/SymfonyFixturesLoader.php
+++ b/Loader/SymfonyFixturesLoader.php
@@ -19,6 +19,7 @@ use function array_key_exists;
 use function array_values;
 use function get_class;
 use function sprintf;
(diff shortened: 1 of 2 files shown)


franmomu (approved):

franmomu:
thanks @GromNaN!

PR #805, opened 2023-12-12
Remove dependency to `jean85/pretty-package-versions`

GromNaN:
Thanks to Composer API v2, we have an easy solution to retrieve the version of a package that is installed.


malarzm (approved):
Cool :ok_hand:

malarzm:
Thanks @GromNaN! Will you be making similar adjustment to ODM itself? I think we were using same package there as well

GromNaN:
Yes, I'd like to do the same for other projects : ODM and mongodb library.

PR #804, opened 2023-12-11
Drop support for Doctrine cache adapter for metadata

GromNaN:
Remove BC layer introduced in #690 to keep support for Doctrine cache adapters. PSR-6 is now required.

PR #810, opened 2023-12-13
Migrate config to PHP

### LINKED ISSUE #807: Move internal Container configuration from XML to PHP
Same as https://github.com/symfony/symfony/issues/37186. The PHP configuration for service definition remove the dependency to ext-xml and make navigating in the code easier.

franmomu:
Closes #807

I've made some changes that we need to see if they are fine. 
- I've split the file because if was easier to check.
- Removed unused services (I haven't found any usage and seem old):
  - `doctrine_mongodb.odm.cache` as alias of `doctrine_mongodb.odm.cache.array`
  - `doctrine_mongodb.odm.cache.array`
- Removed some parameters ending in `.class` and use the class directly in the service as second argument.
- Change these alias:
```
<service id="%doctrine_mongodb.odm.document_manager.class%" alias="doctrine_mongodb.odm.document_manager" public="false" />
<service id="%doctrine_mongodb.odm.class%" alias="doctrine_mongodb" public="false" />
```
I'm not sure if that can be done in PHP and I've changed them to (I guess we can change them in the extension resolving the parameter, but not sure if that's necessary):
```
->alias(DocumentManager::class, 'doctrine_mongodb.odm.document_manager')
->alias(ManagerRegistry::class, 'doctrine_mongodb')
```


Resources/config/mongodb.php
```diff
@@ -0,0 +1,91 @@
...
+        ->set('doctrine_mongodb.odm.cache.apc.class', 'Doctrine\Common\Cache\ApcCache')
+        ->set('doctrine_mongodb.odm.cache.apcu.class', 'Doctrine\Common\Cache\ApcuCache')
+        ->set('doctrine_mongodb.odm.cache.memcache.class', 'Doctrine\Common\Cache\MemcacheCache')
+        ->set('doctrine_mongodb.odm.cache.memcache_host', 'localhost')
+        ->set('doctrine_mongodb.odm.cache.memcache_port', 11211)
+        ->set('doctrine_mongodb.odm.cache.memcache_instance.class', 'Memcache')
+        ->set('doctrine_mongodb.odm.cache.xcache.class', 'Doctrine\Common\Cache\XcacheCache')
+        ->set('doctrine_mongodb.odm.connection.class', Client::class)
+        ->set('doctrine_mongodb.odm.configuration.class', Configuration::class)
+        ->set('doctrine_mongodb.odm.document_manager.class', DocumentManager::class)
```
GromNaN: This parameter should be deprecated. If someone really wants to change the class for this service, they should update its definition in a compiler pass.

I propose to deprecate updating this parameters in 4.7

Resources/config/mongodb.php
```diff
@@ -0,0 +1,91 @@
...
+                service('service_container'),
+            ])
+
+        ->set('doctrine_mongodb.odm.container_repository_factory', ContainerRepositoryFactory::class)
+            ->args([''])
+
+        ->set('doctrine_mongodb.odm.manager_configurator.abstract', ManagerConfigurator::class)
+            ->abstract(true)
+            ->args([
+                [],
```
GromNaN: This argument is always replaced by the extension. You can use `abstract_arg`.

```suggestion
                abstract_arg('enabled filters'),
```

Resources/config/mongodb.php
```diff
@@ -0,0 +1,91 @@
...
+                GhostObjectInterface::class,
+                service('service_container'),
+            ])
+
+        ->set('doctrine_mongodb.odm.listeners.resolve_target_document', ResolveTargetDocumentListener::class)
+
+        ->set('doctrine_mongodb.odm.symfony.fixtures.loader', SymfonyFixturesLoader::class)
+            ->args([
+                service('service_container'),
+            ]);
```
GromNaN: The container injection was removed in [#802](https://github.com/doctrine/DoctrineMongoDBBundle/pull/802/files#diff-80a4f7479361b0ee5647b22762859618d9cd88b06d7fb9167fee7662d971507cL187)
```suggestion
        ->set('doctrine_mongodb.odm.symfony.fixtures.loader', SymfonyFixturesLoader::class);
```

GromNaN (commented):
Splitting the file is a very good idea.

We should get rid of all the `.class` parameters (see https://github.com/symfony/symfony/issues/11881)

DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -17,8 +17,10 @@
 use Doctrine\Common\Cache\RedisCache;
 use Doctrine\Common\DataFixtures\Loader as DataFixturesLoader;
 use Doctrine\Common\EventSubscriber;
+use Doctrine\ODM\MongoDB\Configuration as MongoDBConfiguration;
```
GromNaN: I'm nitpicking, but it's all mongodb here. We can specify ODM.
```suggestion
use Doctrine\ODM\MongoDB\Configuration as ODMConfiguration;
```

GromNaN (approved):
SGTM

GromNaN (approved):

Resources/config/mongodb.php
```diff
@@ -0,0 +1,80 @@
...
+        ->set('doctrine_mongodb.odm.cache.apc.class', 'Doctrine\Common\Cache\ApcCache')
+        ->set('doctrine_mongodb.odm.cache.apcu.class', 'Doctrine\Common\Cache\ApcuCache')
+        ->set('doctrine_mongodb.odm.cache.memcache.class', 'Doctrine\Common\Cache\MemcacheCache')
+        ->set('doctrine_mongodb.odm.cache.memcache_host', 'localhost')
+        ->set('doctrine_mongodb.odm.cache.memcache_port', 11211)
+        ->set('doctrine_mongodb.odm.cache.memcache_instance.class', 'Memcache')
+        ->set('doctrine_mongodb.odm.cache.xcache.class', 'Doctrine\Common\Cache\XcacheCache')
+        ->set('doctrine_mongodb.odm.metadata.driver_chain.class', MappingDriverChain::class)
+        ->set('doctrine_mongodb.odm.metadata.attribute.class', AttributeDriver::class)
+        ->set('doctrine_mongodb.odm.metadata.xml.class', XmlDriver::class);
```
GromNaN, line 34: Related to #811, the parameters should be deprecated here.

franmomu: This is partially related to https://github.com/doctrine/DoctrineMongoDBBundle/pull/804, these parameters are apparently used in the AbstractDoctrineExtension, this is the code related about cache:

https://github.com/symfony/symfony/blob/e0df5c3d9c0281cdfe3885281145e2065b859d32/src/Symfony/Bridge/Doctrine/DependencyInjection/AbstractDoctrineExtension.php#L340-L406

And the related to the metadata:

https://github.com/symfony/symfony/blob/e0df5c3d9c0281cdfe3885281145e2065b859d32/src/Symfony/Bridge/Doctrine/DependencyInjection/AbstractDoctrineExtension.php#L193-L228

GromNaN: Even if we stop using them, we have to keep and deprecate them in 2.x, as they could be used in applications.

franmomu: I was thinking that since these parameters seem to be used for internal configuration, maybe Symfony can try to fetch them first as [build parameters](https://github.com/symfony/symfony/pull/47680) (prefixing the parameter with a ".") and fallback to "normal", I'll see if I have time tomorrow to open a PR there for that.

**Update**: Symfony always calls an abstract method that is implemented in our side to get the parameter name, so the build parameters can be implemented in our side I think.

**Update2:** We can do it for metadata parameters, but for cache parameters I think need to be done in the `AbstractDoctrineExtension`, I'll take a look tomorrow if I find some time.

GromNaN:
Thank you @franmomu

PR #814, opened 2023-12-14
Expose transaction setting

alcaeus:
This complements https://github.com/doctrine/mongodb-odm/pull/2586 by exposing the `use_transactional_flush` setting to the bundle configuration.

### DIFF
--- a/tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
+++ b/tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Doctrine\Bundle\MongoDBBundle\Tests\DependencyInjection;
 
+use Composer\InstalledVersions;
+use Composer\Semver\VersionParser;
 use Doctrine\Bundle\MongoDBBundle\Attribute\MapDocument;
 use Doctrine\Bundle\MongoDBBundle\DependencyInjection\Configuration;
 use Doctrine\Bundle\MongoDBBundle\DependencyInjection\DoctrineMongoDBExtension;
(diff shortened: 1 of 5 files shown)


malarzm (commented):
Configuration reference from docs should be updated in same PR :)

alcaeus:
Updated to target 5.0.x, which we may bump to 5.1.x. I've also added logic to support older ODM versions that don't support transactional flushes. We can decide whether to bump the ODM requirement at a later date.

malarzm (approved):
> which we may bump to 5.1.x

Please do, it's new config switch after all.

alcaeus:
> Please do, it's new config switch after all.

Will do once we have a 5.1.x branch - release automation doesn't handle pre-releases very gracefully, so once 5.0.0 is tagged everything should be in order.

malarzm:
Ah 5.0.0 is still in RC phase 🙈

tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
```diff
@@ -369,4 +372,70 @@ public function testControllerResolver(): void
         $container->compile();
         $this->assertEquals(new MapDocument(null, null, null, [], null, null, null, true), $container->get('controller_resolver_defaults'));
     }
+
+    public function testTransactionalFlushConfigurationWhenNotSupported(): void
+    {
+        if (method_exists(ODMConfiguration::class, 'setUseTransactionalFlush')) {
```
GromNaN: This is a weak test. If the method is renamed, the test is skipped while it should fail.

I think it would be more robust to use `Composer\InstalledVersions`:

```php
Composer\InstalledVersions::satisfies(new VersionParser, 'doctrine/mongodb-odm', '>=2.7@dev')
```

GromNaN (approved):
A suggested improvement, but SGTM as this.

tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
alcaeus: Updated. I wasn't even aware this existed. This is simply lovely :)

GromNaN: I see [3 skipped tests](https://github.com/doctrine/DoctrineMongoDBBundle/actions/runs/7584953100/job/20659752583?pr=814#step:9:15), is the version contraint correct?

PR #821, opened 2023-12-19
Remove deprecated parameters

GromNaN:
Revert deprecation pass from https://github.com/doctrine/DoctrineMongoDBBundle/pull/811 and remove all deprecated parameters.


DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -534,7 +537,12 @@ protected function getMappingResourceExtension(): string
 
     protected function getMetadataDriverClass(string $driverType): string
     {
-        return '%' . $this->getObjectManagerElementName('metadata.' . $driverType . '.class') . '%';
+        return match ($driverType) {
+            'driver_chain' => MappingDriverChain::class,
```
malarzm, line 541: We don't have any consts for these values I presume?

malarzm: Or maybe this should become an enum while we're here? :)

malarzm (approved):

DependencyInjection/DoctrineMongoDBExtension.php
GromNaN: The `getMetadataDriverClass` method is defined and used in the parent class, from doctrine bridge. That's something that needs to be done in the bridge first.

franmomu (approved):

PR #824, opened 2023-12-21
Update Doctrine ODM and Persistence dependencies

GromNaN:
We dropped support for `doctrine/cache` in https://github.com/doctrine/DoctrineMongoDBBundle/pull/804. Applications can upgrade to `doctrine/persistence: 3`.

Updating minimum dependencies helps maintenance by avoiding edge version incompatibility with future releases.


alcaeus (approved):

malarzm (changes requested):
Please evaluate how is ORM and broader ecosystem doing before raising version

GromNaN:
There is already a [huge switch to doctrine/persistence 3](https://packagist.org/packages/doctrine/persistence/stats).
image

This package could be updated: https://github.com/phpstan/phpstan-doctrine/blob/310ab0f1ae9224c3c6aae39fa1e326a4053766c7/composer.json#L28C46-L28C46

Symfony is tested with [persistence 3 only](https://github.com/symfony/symfony/blob/6873ab900f9e5279bcc1ef9347d54b1aaefac74b/composer.json#L41).

malarzm (approved):
Convinced :angel: let's do the same in ODM then

PR #825, opened 2023-12-21
Deprecate `#[AsDocumentListener]` parameters `$method` and `$lazy` (not working)

GromNaN:
If you use the `TestAttributeListener` as it was, in a project, it will throw an exception:

> Attempted to call an undefined method named "prePersist" of class "TestAttributeListener". Did you mean to call "onPrePersist"?

Only the event name is used [when injected into the manager](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Bridge/Doctrine/DependencyInjection/CompilerPass/RegisterEventListenersAndSubscribersPass.php#L118).

`lazy` and `method` have no effect.

### DIFF
--- a/Attribute/AsDocumentListener.php
+++ b/Attribute/AsDocumentListener.php
@@ -6,6 +6,8 @@ namespace Doctrine\Bundle\MongoDBBundle\Attribute;
 
 use Attribute;
 
+use function trigger_deprecation;
+
 /**
  * Service tag to autoconfigure document listeners.
  */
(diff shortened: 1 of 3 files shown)


Attribute/AsDocumentListener.php
```diff
@@ -14,10 +15,28 @@ class AsDocumentListener
...
+        /** @deprecated the method name is the same as the event name */
         public ?string $method = null,
+        /** @deprecated not supported */
         public ?bool $lazy = null,
         public ?string $connection = null,
         public ?int $priority = null,
     ) {
+        // phpcs:disable SlevomatCodingStandard.ControlStructures.EarlyExit.EarlyExitNotUsed
+        if ($method !== null) {
+            Deprecation::trigger(
```
malarzm: We used to use Symfony's deprecation mechanisms in ODM, is it no longer the case?

GromNaN: Great! I'm still learning.

GromNaN: Fixed.

PR #831, opened 2023-12-22
Fix partial service removed when fixtures package is not installed

### LINKED ISSUE #830: Non-existent service after fresh installation on branch 5.0.x
Environment : 
* Symfony 7.0
* PHP 8.3 with mongodb extension

When I install a fresh version of Symfony with the symfony cli :

`symfony new test --webapp --version="stable"`

Then, i add the DoctrineMongoDBBundle :

`composer require doctrine/mongodb-odm-bundle:5.0.x-dev`

Then I answer yes for the recipe execution.

I have the following error while the execution of the @auto-scripts part of composer.json, more precisely in the cache:clear command :

`The service "doctrine_mongodb.odm.command.load_data_fixtures" has a dependency on a non-existent service "doctrine_mongodb.odm.symfony.fixtures.loader".`

I managed to fix this by installing doctrine/data-fixtures but I don't want to have this package in prod environment.

After accessing the page http://127.0.0.1:8000/ after launching `symfony serve` in the test directory, I have the following error :

`The profiler template "@DoctrineMongoDB/Collector/mongodb.html.twig" for data collector "mongodb" does not exist.`

Thanks and mainly thanks for the hard work.

GromNaN:
Fix #830

### DIFF
--- a/src/DependencyInjection/DoctrineMongoDBExtension.php
+++ b/src/DependencyInjection/DoctrineMongoDBExtension.php
@@ -7,7 +7,6 @@ namespace Doctrine\Bundle\MongoDBBundle\DependencyInjection;
 use Composer\InstalledVersions;
 use Doctrine\Bundle\MongoDBBundle\Attribute\AsDocumentListener;
 use Doctrine\Bundle\MongoDBBundle\Attribute\MapDocument;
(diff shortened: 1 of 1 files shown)


malarzm (approved):

franmomu (approved):
After removing the unused `use`.

Issue #536, opened 2019-03-13
Remove usage of bundles from documentation


malarzm:
Leftover from #535

GromNaN:
Done in https://github.com/doctrine/DoctrineMongoDBBundle/pull/834

PR #834, opened 2024-01-04
Add configuration example with `PHP` format

### LINKED ISSUE #806: Add configuration example with `PHP` format
In the [configuration doc](https://github.com/doctrine/DoctrineMongoDBBundle/blob/4.7.x/Resources/doc/config.rst#sample-configuration), the example should be ported to PHP in addition to the current Yaml and Xml examples.

Similar to this:
https://symfony.com/doc/6.4/reference/configuration/framework.html#handler-id

GromNaN:
Fix #806

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -18,7 +18,7 @@ Sample Configuration
             document_managers:
                 default:
                     mappings:
-                        AcmeDemoBundle: ~
+                        App: ~
                     filters:
                         filter-name:
                             class: Class\Example\Filter\ODM\ExampleFilter
@@ -41,7 +41,7 @@ Sample Configuration
                     </doctrine_mongodb:options>
                 </doctrine_mongodb:connection>
                 <doctrine_mongodb:document-manager id="default">
-                    <doctrine_mongodb:mapping name="AcmeDemoBundle" />
+                    <doctrine_mongodb:mapping name="App" />
                     <doctrine_mongodb:filter name="filter-name" enabled="true" class="Class\Example\Filter\ODM\ExampleFilter" />
                     <doctrine_mongodb:metadata-cache-driver type="array" />
                 </doctrine_mongodb:document-manager>
@@ -105,7 +105,7 @@ If you wish to use memcached to cache your metadata, you need to configure the
             document_managers:
                 default:
                     mappings:
-                        AcmeDemoBundle: ~
+                        App: ~
                     metadata_cache_driver:
                         type: memcached
                         class: Symfony\Component\Cache\Adapter\MemcachedAdapter
@@ -125,7 +125,7 @@ If you wish to use memcached to cache your metadata, you need to configure the
 
             <doctrine_mongodb:config default-database="hello_%kernel.environment%">
                 <doctrine_mongodb:document-manager id="default">
-                    <doctrine_mongodb:mapping name="AcmeDemoBundle" />
+                    <doctrine_mongodb:mapping name="App" />
                     <doctrine_mongodb:metadata-cache-driver type="memcached">
                         <doctrine_mongodb:class>Symfony\Component\Cache\Adapter\MemcachedAdapter</doctrine_mongodb:class>
                         <doctrine_mongodb:host>localhost</doctrine_mongodb:host>
@@ -182,9 +182,9 @@ can control. The following configuration options exist for a mapping:
 - ``prefix`` A common namespace prefix that all documents of this mapping
   share. This prefix should never conflict with prefixes of other defined
   mappings otherwise some of your documents cannot be found by Doctrine. This
-  option defaults to the bundle namespace + ``Document``, for example for an
-  application bundle called ``AcmeHelloBundle``, the prefix would be
-  ``Acme\HelloBundle\Document``.
+  option defaults to the application namespace + ``Document``, for example
+  for an application called ``App``, the prefix would be
+  ``App\Document``.
 
 - ``alias`` Doctrine offers a way to alias document namespaces to simpler,
   shorter names to be used in queries or for Repository access.
@@ -219,14 +219,14 @@ The following configuration shows a bunch of mapping examples:
             document_managers:
                 default:
                     mappings:
-                        MyBundle1: ~
-                        MyBundle2: xml
-                        MyBundle3: { type: attribute, dir: Documents/ }
-                        MyBundle4: { type: xml, dir: Resources/config/doctrine/mapping }
-                        MyBundle5:
+                        App: ~
+                        App2: xml
+                        App3: { type: attribute, dir: Documents/ }
+                        App4: { type: xml, dir: config/doctrine/mapping }
+                        App5:
                             type: xml
-                            dir: my-bundle-mappings-dir
-                            alias: BundleAlias
+                            dir: my-app-mappings-dir
+                            alias: AppAlias
                         doctrine_extensions:
                             type: xml
                             dir: "%kernel.project_dir%/src/vendor/DoctrineExtensions/lib/DoctrineExtensions/Documents"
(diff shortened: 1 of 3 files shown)


docs/config.rst
```diff
@@ -214,6 +258,34 @@ The following configuration shows a bunch of mapping examples:
...
+        use Symfony\Config\DoctrineMongodbConfig;
+
+        return static function (DoctrineMongodbConfig $config): void {
+            $config->documentManager('default')
+                ->mapping('MyBundle1')
+                ->mapping('MyBundle2')
+                    ->type('xml')
+                ->mapping('MyBundle3')
+                    ->type('attribute')
+                    ->dir(__DIR__.'/../src/Document')
```
malarzm: I think the directories should be consistent with previous sections, so this should read `Documents/`

docs/config.rst
```diff
@@ -214,6 +258,34 @@ The following configuration shows a bunch of mapping examples:
...
+                ->mapping('MyBundle4')
+                    ->type('xml')
+                    ->dir(__DIR__.'/../config/doctrine/mapping')
```
malarzm: This is missing Resources. Or if `Resources` is no longer encouraged we should update previous examples

docs/config.rst
```diff
@@ -214,6 +258,34 @@ The following configuration shows a bunch of mapping examples:
...
+                ->mapping('MyBundle5')
+                    ->type('xml')
+                    ->dir('my-bundle-mappings-dir')
+                    ->alias('BundleAlias')
```
malarzm: Haven't we removed aliases and entire `Smh:Document` notation along with doctrine/common 3.0 requirement? Or is this something else?

docs/config.rst
```diff
@@ -214,6 +258,34 @@ The following configuration shows a bunch of mapping examples:
...
+                ->mapping('MyBundle6')
+                    ->type('xml')
+                    ->dir('%kernel.project_dir%/src/vendor/DoctrineExtensions/lib/DoctrineExtensions/Documents')
```
malarzm: I see that Symfony provides a `param` function that can be used instead `%notation%`. Shall we endorse the new way or is there no community consensus yet?

docs/config.rst
```diff
@@ -298,6 +381,24 @@ Filters may be registered with a document manager by using the following syntax:
...
         </container>
 
+    .. code-block:: php
+
+        use Symfony\Config\DoctrineMongodbConfig;
+
+        return static function (DoctrineMongodbConfig $config): void {
+            $config->documentManager('default')
+                ->filter('basic_filter')
+                    ->class('Vendor\Filter\BasicFilter')
```
malarzm: We should be encouraging `::class` notation in both `->class` calls

docs/config.rst
```diff
@@ -684,6 +863,69 @@ Full Default Configuration
...
+            $config->autoGenerateHydratorClasses(0);
+            $config->autoGenerateProxyClasses(0);
+            $config->defaultConnection('');
+            $config->defaultDatabase('default');
+            $config->defaultDocumentManager('');
+            $config->hydratorDir('%kernel.cache_dir%/doctrine/odm/mongodb/Hydrators');
+            $config->hydratorNamespace('Hydrators');
+            $config->proxyDir('%kernel.cache_dir%/doctrine/odm/mongodb/Proxies');
+            $config->proxyNamespace('Proxies');
+            $config->fixtureLoader('Symfony\Bridge\Doctrine\DataFixtures\ContainerAwareLoader');
```
malarzm: `::class` please, everywhere

malarzm (commented):
Sorry for the number of comments 🙈

docs/config.rst
GromNaN: Using the constants is one of the advantages of using the PHP configuration, I think it's better to show the full feature as it should be used in the documentation.

docs/config.rst
GromNaN: Yes, I'll use it.

docs/config.rst
```diff
@@ -684,6 +863,69 @@ Full Default Configuration
...
+                ->defaultDocumentRepositoryClass('')
+                ->defaultGridfsRepositoryClass('')
+                ->repositoryFactory('')
+                ->logging(true)
+                ->autoMapping(false)
+                ->metadataCacheDriver()
+                    ->type('')
+                    ->class('')
+                    ->host('')
+                    ->port('')
```
GromNaN: This list of "defaults" is incorrect. There is no default values as this are not set by default.

franmomu (commented):
I remembered that I starting changing references to bundle to fix https://github.com/doctrine/DoctrineMongoDBBundle/issues/536, we should remove them here as well (I've just created https://github.com/doctrine/DoctrineMongoDBBundle/pull/835)

GromNaN:
@franmomu @malarzm The PR have been updated after your reviews.
I updated the first example to take the default configuration provided in the [Symfony recipe.](https://github.com/symfony/recipes-contrib/blob/main/doctrine/mongodb-odm-bundle/4.4/config/packages/doctrine_mongodb.yaml). ~The `is_bundle` part is very important since Bundle structure is no longer the default.~

docs/config.rst
```diff
@@ -18,11 +18,11 @@ Sample Configuration
             document_managers:
                 default:
                     mappings:
-                        AcmeDemoBundle: ~
-                    filters:
-                        filter-name:
-                            class: Class\Example\Filter\ODM\ExampleFilter
-                            enabled: true
+                        App:
+                            is_bundle: false
```
GromNaN, line 22: The `is_bundle` and `alias` configs are not necessary for a default setup. I update the recipe: https://github.com/symfony/recipes-contrib/pull/1582

src/DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -462,14 +462,14 @@ private function normalizeDriverOptions(array $connection): array
      *
      *  doctrine_mongodb:
      *     mappings:
-     *         MyBundle1: ~
-     *         MyBundle2: xml
-     *         MyBundle3: { type: attribute }
-     *         MyBundle4: { type: xml, dir: Resources/config/doctrine/mapping }
-     *         MyBundle5:
+     *         App1: ~
```
GromNaN, line 465: This configs need to be updated, as `dir` is required when outside of a bundle: https://github.com/symfony/symfony/blob/7bbc9be7ea7d607fa901edb2068c2cea13d24978/src/Symfony/Bridge/Doctrine/DependencyInjection/AbstractDoctrineExtension.php#L68

franmomu: I was wondering if we should keep these examples here if we have them in the documentation.

Maybe we can remove them here and update the ones in the documentation adding the `dir` option?

docs/config.rst
```diff
@@ -138,9 +182,9 @@ can control. The following configuration options exist for a mapping:
...
   share. This prefix should never conflict with prefixes of other defined
   mappings otherwise some of your documents cannot be found by Doctrine. This
-  option defaults to the bundle namespace + ``Document``, for example for an
-  application bundle called ``AcmeHelloBundle``, the prefix would be
-  ``Acme\HelloBundle\Document``.
+  option defaults to the application namespace + ``Document``, for example
+  for an application called ``App``, the prefix would be
+  ``App\Document``.
 
 - ``alias`` Doctrine offers a way to alias document namespaces to simpler,
```
malarzm, line 189: I still think we've missed deprecating this one :)

malarzm (approved):
Thanks for the changes! 🍻

src/DependencyInjection/DoctrineMongoDBExtension.php
GromNaN: Having an example for attribute and xml, and for a bundle would be enough.

docs/config.rst
GromNaN: Let's deprecate it for 5.1 and remove it from the doc.

GromNaN:
Status: I need to check the examples. I'm not sure everything is correct and optimal.

IonBazan (approved):

PR #853, opened 2024-09-18
Support custom not found message based on Doctrine Bridge MapEntity

Avariya:
Use custom not found message from ORM bundle. https://github.com/symfony/symfony/pull/51562

### DIFF
--- a/src/Attribute/MapDocument.php
+++ b/src/Attribute/MapDocument.php
@@ -6,7 +6,11 @@ namespace Doctrine\Bundle\MongoDBBundle\Attribute;
 
 use Attribute;
 use Doctrine\Bundle\MongoDBBundle\ArgumentResolver\DocumentValueResolver;
+use RuntimeException;
 use Symfony\Bridge\Doctrine\Attribute\MapEntity;
+
+use function is_string;
+use function property_exists;
 
 #[Attribute(Attribute::TARGET_PARAMETER)]
 class MapDocument extends MapEntity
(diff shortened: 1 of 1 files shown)


Avariya:
Tricky part is dependency bump, as 6.4 and 7.0 are not supported. Probably it should be new major version of [DoctrineMongoDBBundle](https://github.com/doctrine/DoctrineMongoDBBundle)?

src/Attribute/MapDocument.php
```diff
@@ -21,7 +21,8 @@ public function __construct(
         public array|string|null $id = null,
         bool $disabled = false,
         string $resolver = DocumentValueResolver::class,
+        public ?string $message = null,
     ) {
-        parent::__construct($class, $objectManager, $expr, $mapping, $exclude, $stripNull, $id, null, $disabled, $resolver);
+        parent::__construct($class, $objectManager, $expr, $mapping, $exclude, $stripNull, $id, null, $disabled, $resolver, $message);
```
IonBazan, line 42: I think we can avoid locking in `^7.1` by checking if `$message` property exists in base class. We could also trigger a warning if someone passed the message when using older version of Symfony.

Avariya: Done. 
Not sure should it be `@trigger_error` or `RuntimeException` as trigger_error newer used at the project.

GromNaN: `RuntimeException` is correct. Setting the `$message` argument was not possible previously, so nothing is being broken.

GromNaN (commented):
The sole purpose of this constructor is to change the default `$resolver`. We should try to add a protected constant `DEFAULT_RESOLVER` to the parent class to remove this constructor here.

GromNaN (approved):

SenseException (approved):

GromNaN:
Thank you @Avariya

PR #865, opened 2024-11-29
Use AbstractLogger instead of Closure in LoadDataFixturesDoctrineODMCommand.

mickverm:
Fixes https://github.com/doctrine/DoctrineMongoDBBundle/issues/864

PR #866, opened 2024-11-29
Restrict compatibility constraints with `psr/log` and `doctrine/data-fixtures`

GromNaN:
- Change in https://github.com/doctrine/DoctrineMongoDBBundle/pull/865 is not compatible with `psr/log:^1`
- We cannot garantee compatibility with `doctrine/data-fixtures:>3`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "composer-runtime-api": "^2.0",
         "doctrine/mongodb-odm": "^2.6",
         "doctrine/persistence": "^3.0",
-        "psr/log": "^1.0 || ^2.0 || ^3.0",
+        "psr/log": "^2.0 || ^3.0",
         "symfony/config": "^6.4 || ^7.0",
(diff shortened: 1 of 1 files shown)


mickverm (approved):

composer.json
```diff
@@ -29,7 +29,7 @@
         "composer-runtime-api": "^2.0",
         "doctrine/mongodb-odm": "^2.6",
         "doctrine/persistence": "^3.0",
-        "psr/log": "^1.0 || ^2.0 || ^3.0",
+        "psr/log": "^2.0 || ^3.0",
```
GromNaN, line 32: Reverted during [upmerge](https://github.com/doctrine/DoctrineMongoDBBundle/pull/873/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R32) by relaxing types.

PR #876, opened 2025-01-06
Exclude classes with `#[Document]` attribute from container

GromNaN:
Related to https://github.com/symfony/symfony/discussions/59280

Avoids any problems if the `src/Document` directory is not excluded from the container configuration. This directory is not excluded by default by the Symfony recipe.


SenseException (approved):
It can be seen as a bug, so my guess is that the 5.2.x branch is the right choice, but I'd like a second reviewer about that.

GromNaN:
In my opinion, this is a new feature that simplifies configuration.

jmikola (approved):

PR #877, opened 2025-01-22
Use LazyGhostObject for proxy objects

GromNaN:
Requires the release of `doctrine/mongodb-odm: 2.10` with LazyGhostObject support provided by `symfony/var-exporter`: 

Enable `UseLazyGhostObject` configuration and fix proxy interface after the default configuration is reset to Proxy Manager by https://github.com/doctrine/mongodb-odm/pull/2720 for compatibility with older versions of the bundle.

New configuration setting: `enable_lazy_ghost_objects` which is true by default when the feature is available.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "php": "^8.1",
         "ext-mongodb": "^1.16",
         "composer-runtime-api": "^2.0",
-        "doctrine/mongodb-odm": "^2.10@dev",
+        "doctrine/mongodb-odm": "^2.6",
         "doctrine/persistence": "^3.0",
         "psr/log": "^1.0 || ^2.0 || ^3.0",
         "symfony/config": "^6.4 || ^7.0",
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "php": "^8.1",
         "ext-mongodb": "^1.16",
         "composer-runtime-api": "^2.0",
-        "doctrine/mongodb-odm": "^2.6",
+        "doctrine/mongodb-odm": "^2.10@dev",
         "doctrine/persistence": "^3.0",
         "psr/log": "^1.0 || ^2.0 || ^3.0",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -27,7 +27,7 @@
         "php": "^8.1",
         "ext-mongodb": "^1.16",
         "composer-runtime-api": "^2.0",
-        "doctrine/mongodb-odm": "^2.6",
+        "doctrine/mongodb-odm": "^2.10@dev",
```
GromNaN: Update after the version is released.

```suggestion
        "doctrine/mongodb-odm": "^2.10",
```

alcaeus:
LGTM in general. Do we want to give people the opportunity to disable lazy ghost objects in the config (and default to `true` with the setting), or would that complicate things considering we're passing the interface name to the ManagerRegistry?

composer.json
GromNaN: I reverted the required update. The lazy ghost object option is enabled only when available.

GromNaN:
I added a configuration that is true when the feature is available, and reverted the need to update minimal version of `doctrine/mongodb-odm`.

src/DependencyInjection/Configuration.php
```diff
@@ -39,6 +40,13 @@ public function getConfigTreeBuilder(): TreeBuilder
             ->children()
                 ->scalarNode('proxy_namespace')->defaultValue('MongoDBODMProxies')->end()
                 ->scalarNode('proxy_dir')->defaultValue('%kernel.cache_dir%/doctrine/odm/mongodb/Proxies')->end()
+                ->booleanNode('enable_lazy_ghost_objects')
```
GromNaN, line 43: Same name used in the ORM: https://github.com/doctrine/DoctrineBundle/blob/2363c43d9815a11657e452625cd64172d5587486/src/DependencyInjection/Configuration.php#L512

alcaeus (approved):

PR #890, opened 2025-05-26
Fix Ubuntu version in GitHub Action

GromNaN:
`ubuntu-20.04` runners are no longer available. Using `ubuntu-24.04` instead.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -21,7 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - "ubuntu-20.04"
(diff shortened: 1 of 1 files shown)


GromNaN:
Do I need to backport this fix into [4.7.x](https://github.com/doctrine/DoctrineMongoDBBundle/tree/4.7.x)? We did not update this branch for 2 years. I'm not sure if we consider this branch is still maintained.

xabbuh (approved):

SenseException (approved):
[4.7. isn't maintained anymore](https://www.doctrine-project.org/projects/doctrine-mongodb-bundle.html)

PR #889, opened 2025-05-26
[Encryption] Add `autoEncryption` configuration to the client

GromNaN:
Introduce the configuration of the [`autoEncryption` driver option](https://www.php.net/manual/en/mongodb-driver-manager.construct.php) in order to support automatic Client Side Field Level Encryption.

This requires a MongoDB Atlas or Entreprise cluster.

PR #897, opened 2025-06-17
Enable auto encryption

GromNaN:
Integration for the Queryable Encryption feature of MongoDB ODM: https://github.com/doctrine/mongodb-odm/pull/2779

- [x] https://github.com/doctrine/DoctrineMongoDBBundle/pull/889
- [x] https://github.com/doctrine/DoctrineMongoDBBundle/pull/899
- [x] Diagnostic command https://github.com/doctrine/DoctrineMongoDBBundle/pull/898
- [x] Inject default `masterKey` and `kmsProvider` name into the ODM configuration for the Schema Manager. https://github.com/doctrine/DoctrineMongoDBBundle/pull/901
- [x] ~https://github.com/doctrine/DoctrineMongoDBBundle/pull/902~
- [x] https://github.com/doctrine/DoctrineMongoDBBundle/pull/905
- [x] https://github.com/doctrine/DoctrineMongoDBBundle/pull/912
- [x] https://github.com/doctrine/DoctrineMongoDBBundle/pull/913
- [x] https://github.com/doctrine/DoctrineMongoDBBundle/pull/914
- [x] https://github.com/doctrine/DoctrineMongoDBBundle/pull/915
- [x] Revert composer version constraint on `doctrine/mongodb-odm: dev-feature/queryable-encryption` branch


docs/config.rst
```diff
@@ -624,6 +624,10 @@ Otherwise you will get a *auth failed* exception.
                 ]);
         };
 
+Using Queryable Encryption
+--------------------------
+
+TODO: Add documentation for queryable encryption configuration.
```
alcaeus: @GromNaN should we add something here or link to the ODM docs?

alcaeus (commented):
One question about docs, but code changes LGTM!

composer.json
```diff
@@ -25,9 +25,9 @@
     "homepage": "http://www.doctrine-project.org",
     "require": {
         "php": "^8.1",
-        "ext-mongodb": "^1.16 || ^2",
+        "ext-mongodb": "^1.21 || ^2",
         "composer-runtime-api": "^2.0",
-        "doctrine/mongodb-odm": "^2.6",
+        "doctrine/mongodb-odm": "dev-feature/queryable-encryption as 2.12.x-dev",
```
jmikola: Noted that this PR depends on https://github.com/doctrine/mongodb-odm/pull/2779

config/schema/mongodb-1.0.xsd
```diff
@@ -119,6 +123,82 @@
     <xsd:attribute name="value" type="xsd:string" use="required" />
   </xsd:complexType>
 
+  <xsd:complexType name="auto-encryption">
```
jmikola, line 126: I understand you're attempting to model all of the [`autoEncryption` driver options](https://www.php.net/manual/en/mongodb-driver-manager.construct.php#mongodb-driver-manager.construct-autoencryption) here.

Is `keyVaultClient` the only intentional omission? That option takes a `MongoDB\Driver\Manager` instance, which wouldn't be possible to express here unless you implemented some sort of registry and translated string identifiers. 

In the DI configuration class, it looks like you accept a service ID as a string and convert it to a reference. Could that not be done here?

src/DependencyInjection/Configuration.php
```diff
@@ -339,6 +342,133 @@ private function addConnectionsSection(ArrayNodeDefinition $rootNode): void
...
+                                            ->scalarNode('keyName')->end()
+                                            ->scalarNode('keyVersion')->end()
+                                            // GCP
+                                            ->scalarNode('email')->end()
+                                            ->scalarNode('privateKey')->end()
+                                            ->scalarNode('endpoint')->end()
+                                            ->scalarNode('projectId')->end()
+                                            ->scalarNode('location')->end()
+                                            ->scalarNode('keyRing')->end()
+                                            //->scalarNode('keyName')->end()
```
jmikola, line 388: I assume these are commented out because the nodes are already defined for another KMS provider. I'd suggest adding a note for that, or maybe replacing the contents of these comments with something like:

```
// scalarNode('keyVersion') was already defined
```

src/DependencyInjection/Configuration.php
```diff
@@ -339,6 +342,133 @@ private function addConnectionsSection(ArrayNodeDefinition $rootNode): void
...
+                                                    ->variableNode('keyId')->isRequired()->cannotBeEmpty()->end()
+                                                    ->arrayNode('queries')
+                                                        ->children()
+                                                            ->scalarNode('queryType')->isRequired()->cannotBeEmpty()->end()
+                                                            ->variableNode('min')->end()
+                                                            ->variableNode('max')->end()
+                                                            ->integerNode('sparsity')->end()
+                                                            ->integerNode('precision')->end()
+                                                            ->integerNode('trimFactor')->end()
+                                                            ->integerNode('contention')->end()
```
jmikola: Noted that these definitions correspond to https://www.mongodb.com/docs/manual/core/queryable-encryption/fundamentals/encrypt-and-query/

src/DependencyInjection/Configuration.php
```diff
@@ -339,6 +342,133 @@ private function addConnectionsSection(ArrayNodeDefinition $rootNode): void
...
+                                            ->scalarNode('tlsClientCertificateKeyFile')->end()
+                                            ->scalarNode('tlsClientCertificateKeyFilePassword')->end()
+                                            // Local
+                                            ->scalarNode('key')->end()
+                                        ->end()
+                                    ->end()
+                                    ->arrayNode('schemaMap')
+                                        ->prototype('variable')->end()
+                                    ->end()
+                                    ->arrayNode('encryptedFieldsMap')
```
jmikola, line 399: The [Manager constructor docs](https://www.php.net/manual/en/mongodb-driver-manager.construct.php#mongodb-driver-manager.construct-autoencryption) define this as a map of namespaces to `encryptedFields` objects.

Looking at the example in [`Database::createEncryptedCollection`](
https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBDatabase-createEncryptedCollection/#example), I wonder if we're missing a `fields` property here. Or is that covered by the sequence of two `prototype('array')` calls below?

config/schema/mongodb-1.0.xsd
```diff
@@ -119,6 +123,82 @@
...
+    <xsd:sequence>
+      <xsd:element name="kmsProvider" type="kms-provider" minOccurs="0" maxOccurs="1" />
```
jmikola: It's possible to define multiple providers and at least one is required. Should this be changed to `minOccurs="1"` with no upper bound? I suppose the actual upper bound would be the number of distinct provider types but I don't think we'd need to enforce that here.

src/DataCollector/ConnectionDiagnostic.php
```diff
@@ -0,0 +1,96 @@
...
+class ConnectionDiagnostic
+{
+    private const CLIENT_ENCRYPTION_OPTION_NAMES = [
+        'keyVaultClient',
+        'keyVaultNamespace',
+        'kmsProviders',
+        'tlsOptions',
+    ];
+
+    private const SUPPORTED_SERVER_TYPES = [
```
jmikola: Is this class only used for CSFLE? If so, should the name reflect that? ConnectionDiagnostic sounds more generic.

src/DataCollector/ConnectionDiagnostic.php
```diff
@@ -0,0 +1,96 @@
...
+        $server    = $this->client->getManager()->selectServer(new ReadPreference(ReadPreference::PRIMARY_PREFERRED));
+        $buildInfo = $server->executeCommand('admin', new Command(['buildInfo' => 1]))->toArray()[0] ?? null;
+
+        $version = $buildInfo->version ?? null;
+
+        return [
+            'topologyName' => $this->getTopologyType($server),
+            'topologySupported' => in_array($server->getType(), self::SUPPORTED_SERVER_TYPES),
+            'version' => $version,
+            'versionSupported' => $version ? version_compare($version, '8.0.0', '>=') : false,
```
jmikola: Why is 8.0.0 used here? [Queryable Encryption Compatibility](https://www.mongodb.com/docs/manual/core/queryable-encryption/reference/compatibility/#queryable-encryption-compatibility) suggests that QE requires 7.0, and [CSFLE goes back further](https://www.mongodb.com/docs/manual/core/queryable-encryption/reference/compatibility/#client-side-field-level-encryption-compatibility).

Separate question: would you also need to check that libmongocrypt is enabled in the extension and that mongocryptd and/or crypt_shared is available, as we do in PHPLIB tests?

src/DataCollector/ConnectionDiagnostic.php
```diff
@@ -0,0 +1,96 @@
...
+
+    public function usesAutoEncryption(): bool
+    {
+        return isset($this->driverOptions['autoEncryption']);
+    }
+
+    private function getTopologyType(Server $server): string
+    {
+        return match ($server->getType()) {
+            Server::TYPE_STANDALONE => 'Standalone',
```
jmikola, line 81: You may also want to check for a [load balancer](https://www.php.net/manual/en/class.mongodb-driver-server.php#mongodb-driver-server.constants.type-load-balancer), for which you can assume the backing MongoDB service is a sharded cluster.

On a separate note, it may make more sense to determine QE compatibility by rejecting a standalone and accepting all other server types. That's what we do in PHPLIB's test suite.

src/DataCollector/EncryptionDiagnostic.php
```diff
@@ -0,0 +1,106 @@
...
+
+        return null;
+    }
+
+    private function getExtensionInfo(): string
+    {
+        $extension = new ReflectionExtension('mongodb');
+
+        ob_start();
+        $extension->info();
```
jmikola, line 89: I see [`ReflectionExtension::export()`](https://www.php.net/manual/en/reflectionextension.export.php) was removed. Presumably that would mean we wouldn't need output buffering.

src/DataCollector/EncryptionDiagnostic.php
```diff
@@ -0,0 +1,106 @@
...
+use function ob_get_contents;
+use function ob_start;
+use function phpversion;
+use function preg_match;
+use function preg_quote;
+use function sprintf;
+use function trim;
+
+/** @internal */
+class EncryptionDiagnostic
```
jmikola, line 26: Would this be useful to include in PHPLIB directly? I can imagine you might want to utilize this in the Laravel integration at some point.

src/DependencyInjection/Configuration.php
```diff
@@ -339,6 +342,133 @@ private function addConnectionsSection(ArrayNodeDefinition $rootNode): void
                                     ->end()
                                 ->end()
                             ->end()
+                            ->arrayNode('autoEncryption')
+                                ->children()
+                                    ->booleanNode('bypassAutoEncryption')->end()
+                                    ->scalarNode('keyVaultClient')->end()
```
jmikola, line 348: Noted that `scalarNode()` is accepted here because you later convert this service ID into a Reference.

src/DependencyInjection/Configuration.php
```diff
@@ -339,6 +342,133 @@ private function addConnectionsSection(ArrayNodeDefinition $rootNode): void
...
+                                    ->scalarNode('keyVaultNamespace')
+                                        ->validate()
+                                            ->ifTrue(static fn ($v) => ! preg_match('/^.+\..+$/', $v))
+                                            ->thenInvalid('Invalid keyVaultNamespace format. It should be "database.collection".')
+                                        ->end()
+                                    ->end()
+                                    ->arrayNode('masterKey')
+                                        ->prototype('variable')->end()
+                                    ->end()
+                                    ->arrayNode('kmsProvider')
```
jmikola, line 358: Related to my comment in the XSD file, multiple KMS providers should be accepted.

src/DependencyInjection/Configuration.php
```diff
@@ -339,6 +342,133 @@ private function addConnectionsSection(ArrayNodeDefinition $rootNode): void
...
+                                            ->scalarNode('projectId')->end()
+                                            ->scalarNode('location')->end()
+                                            ->scalarNode('keyRing')->end()
+                                            //->scalarNode('keyName')->end()
+                                            //->scalarNode('keyVersion')->end()
+                                            // KMIP
+                                            //->scalarNode('endpoint')->end()
+                                            ->scalarNode('tlsCAFile')->end()
+                                            ->scalarNode('tlsClientCertificateKeyFile')->end()
+                                            ->scalarNode('tlsClientCertificateKeyFilePassword')->end()
```
jmikola: Was `tlsDisableOCSPEndpointCheck` intentionally excluded? I know there was some talk about deprecating OCSP, but I wasn't sure.

src/DependencyInjection/Configuration.php
```diff
@@ -339,6 +342,133 @@ private function addConnectionsSection(ArrayNodeDefinition $rootNode): void
...
+                                            ->scalarNode('privateKey')->end()
+                                            ->scalarNode('endpoint')->end()
+                                            ->scalarNode('projectId')->end()
+                                            ->scalarNode('location')->end()
+                                            ->scalarNode('keyRing')->end()
+                                            //->scalarNode('keyName')->end()
+                                            //->scalarNode('keyVersion')->end()
+                                            // KMIP
+                                            //->scalarNode('endpoint')->end()
+                                            ->scalarNode('tlsCAFile')->end()
```
jmikola: Why are these TLS options included under the `kmsProvider` node? I think they should be defined in `autoEncryption.tlsOptions`.

docs/encryption.rst
```diff
@@ -0,0 +1,293 @@
...
+                    ],
+                    // ... other options ...
+                ]);
+        };
+
+Supported KMS Providers
+-----------------------
+
+The ``kmsProvider`` option specifies a single KMS provider that will be used for encryption.
+The type of KMS provider is specified with the ``type`` property along with its options.
```
jmikola, line 56: Do you intend to support automatic options fetched from environment variables? In that case, the KMS provider would be expressed as an empty object when provided to the extension.

This is mentioned in the [`Manager::__construct()`](https://www.php.net/manual/en/mongodb-driver-manager.construct.php#mongodb-driver-manager.construct-autoencryption) docs and also in [`Manager::createClientEncryption()`](https://www.php.net/manual/en/mongodb-driver-manager.createclientencryption.php).

docs/encryption.rst
```diff
@@ -0,0 +1,293 @@
...
+
+Encrypted Fields Map
+--------------------
+
+You can configure which fields are encrypted in each collection by specifying the
+``autoEncryption.encryptedFieldsMap`` option in the connection configuration.
+This setting is **recommended** for improved security and performance.
+
+- If the connection ``encryptedFieldsMap`` object contains a key for the specified
+  collection, the client uses that object to perform automatic Queryable Encryption,
```
jmikola: Would the key here actually be the collection's full namespace (e.g. `db.coll`)?

docs/encryption.rst
```diff
@@ -0,0 +1,293 @@
...
+  rather than using the remote schema. At minimum, the local rules must encrypt
+  all fields that the remote schema does.
+
+- If the connection ``encryptedFieldsMap`` object doesn't contain a key for the
+  specified collection, the client downloads the server-side remote schema for
```
jmikola: Instead of saying "downloads", consider saying "the client will query the server for the collection schema." This would be a good location to reiterate why a client-defined schema is more reliable.

docs/encryption.rst
```diff
@@ -0,0 +1,293 @@
...
+  the collection and uses it instead.
+
+For more details, see the official MongoDB documentation:
+`Encrypted Fields and Enabled Queries <https://www.mongodb.com/docs/manual/core/queryable-encryption/fundamentals/encrypt-and-query/>`_.
+
+Note that there is no ``fields`` key in the configuration of each collection
```
jmikola: I missed this when leaving an earlier comment. Was this done just to be more concise?

Looking at [Create an Encryption Schema](https://www.mongodb.com/docs/manual/core/queryable-encryption/qe-create-encryption-schema/#std-label-qe-create-encryption-schema), `fields` does appear to be the only top-level key. But I'd worry that skipping it here might back you into a corner if another top-level key was introduced down the line.

src/DependencyInjection/Configuration.php
```diff
@@ -339,6 +342,133 @@ private function addConnectionsSection(ArrayNodeDefinition $rootNode): void
...
+                                    ->end()
+                                    ->arrayNode('schemaMap')
+                                        ->prototype('variable')->end()
+                                    ->end()
+                                    ->arrayNode('encryptedFieldsMap')
+                                        ->useAttributeAsKey('name', false)
+                                        ->beforeNormalization()
+                                            ->always(static function ($v) {
+                                                if (is_string($v)) {
+                                                    return json_decode($v, true, 512, JSON_THROW_ON_ERROR);
```
jmikola, line 407: In the RST docs, you present an example of using extended JSON (with a `$binary` object). In that case, `json_decode()` will not yield the desired `MongoDB\BSON\Binary` object. You'd need to use the extension's functions/classes to parse extended JSON.

config/schema/mongodb-1.0.xsd
```diff
@@ -119,6 +123,82 @@
...
+    <xsd:attribute name="projectId" type="xsd:string" use="optional" />
+    <xsd:attribute name="location" type="xsd:string" use="optional" />
+    <xsd:attribute name="keyRing" type="xsd:string" use="optional" />
+    <!-- <xsd:attribute name="keyName" type="xsd:string" use="optional" /> -->
+    <!-- <xsd:attribute name="keyVersion" type="xsd:string" use="optional" /> -->
+    <!-- KMIP -->
+    <!-- <xsd:attribute name="endpoint" type="xsd:string" use="optional" /> -->
+    <xsd:attribute name="tlsCAFile" type="xsd:string" use="optional" />
+    <xsd:attribute name="tlsClientCertificateKeyFile" type="xsd:string" use="optional" />
+    <xsd:attribute name="tlsClientCertificateKeyFilePassword" type="xsd:string" use="optional" />
```
jmikola: Related to my comment on the configuration, I don't think any KMS provider has these TLS options.

config/schema/mongodb-1.0.xsd
```diff
@@ -119,6 +123,82 @@
...
+    <xsd:attribute name="sparsity" type="xsd:string" use="optional" />
+    <xsd:attribute name="trimFactor" type="xsd:string" use="optional" />
+  </xsd:complexType>
+
+  <xsd:complexType name="extra-options">
+    <xsd:attribute name="mongocryptdURI" type="xsd:string" use="optional" />
+    <xsd:attribute name="mongocryptdBypassSpawn" type="xsd:boolean" use="optional" />
+    <xsd:attribute name="mongocryptdSpawnPath" type="xsd:string" use="optional" />
+    <xsd:attribute name="mongocryptdSpawnArgs" type="xsd:string" use="optional" />
+    <xsd:attribute name="cryptSharedLibPath" type="xsd:string" use="optional" />
```
jmikola: The `cryptSharedLibRequired` boolean option is omitted.

src/DependencyInjection/Configuration.php
```diff
@@ -339,6 +342,133 @@ private function addConnectionsSection(ArrayNodeDefinition $rootNode): void
...
+                                ->children()
+                                    ->booleanNode('bypassAutoEncryption')->end()
+                                    ->scalarNode('keyVaultClient')->end()
+                                    ->scalarNode('keyVaultNamespace')
+                                        ->validate()
+                                            ->ifTrue(static fn ($v) => ! preg_match('/^.+\..+$/', $v))
+                                            ->thenInvalid('Invalid keyVaultNamespace format. It should be "database.collection".')
+                                        ->end()
+                                    ->end()
+                                    ->arrayNode('masterKey')
```
jmikola, line 355: Why is the `masterKey` defined under `autoEncryption` here and in XML?

Do I understand correctly that this is only used for `Database::createEncryptedCollection()`? In that case I wonder if it's something that should instead be configured at the model level (per collection), with this possibly renamed to `defaultMasterKey`.

src/DependencyInjection/Configuration.php
```diff
@@ -339,6 +342,133 @@ private function addConnectionsSection(ArrayNodeDefinition $rootNode): void
...
+                                                            ->integerNode('trimFactor')->end()
+                                                            ->integerNode('contention')->end()
+                                                        ->end()
+                                                    ->end()
+                                                ->end()
+                                            ->end()
+                                        ->end()
+                                    ->end()
+                                    ->arrayNode('extraOptions')
+                                        ->prototype('variable')->end()
```
jmikola: You modeled these options in the XSD file. Any reason you chose not to do so here?

src/DependencyInjection/Configuration.php
```diff
@@ -339,6 +342,133 @@ private function addConnectionsSection(ArrayNodeDefinition $rootNode): void
...
+                                    ->end()
+                                    ->booleanNode('bypassQueryAnalysis')->end()
```
jmikola, line 453: Did you also want to add `bypassAutoEncryption`? That's an older option, so I don't think it's directly relevant to QE.

src/DependencyInjection/Configuration.php
```diff
@@ -339,6 +342,133 @@ private function addConnectionsSection(ArrayNodeDefinition $rootNode): void
...
+                                    ->arrayNode('tlsOptions')
+                                        ->children()
+                                            ->scalarNode('tlsCAFile')->end()
+                                            ->scalarNode('tlsCertificateKeyFile')->end()
+                                            ->scalarNode('tlsCertificateKeyFilePassword')->end()
+                                            ->booleanNode('tlsAllowInvalidCertificates')->end()
+                                            ->booleanNode('tlsAllowInvalidHostnames')->end()
+                                            ->booleanNode('tlsDisableCertificateRevocationCheck')->end()
+                                            ->booleanNode('tlsDisableOCSPEndpointCheck')->end()
+                                            ->booleanNode('tlsInsecure')->end()
```
jmikola: This looks a mix of the driver-level TLS options and the subset supported by `autoEncryption.tlsOptions`. I noted that they're used to extend the KMS providers, so that might explain why I saw some of these in the earlier KMS provider definitions.

config/schema/mongodb-1.0.xsd
jmikola: If the bundle will intentionally only support a single KMS provider, let's just make sure that difference is called out in the documentation.

src/DependencyInjection/Configuration.php
jmikola: Noted that the bundle intentionally takes only a single provider.

docs/config.rst
```diff
@@ -699,6 +703,33 @@ Full Default Configuration
...
+                            tlsOptions:                   # TLS options for the Key Vault client (if keyVaultClient is not specified)
+                                tlsCAFile:                              null  # Path to CA file, e.g., /path/to/key-vault-ca.pem
+                                tlsCertificateKeyFile:                  null  # Path to client cert/key file, e.g., /path/to/key-vault-client.pem
+                                tlsCertificateKeyFilePassword:          null  # Password for client cert/key file
+                                tlsAllowInvalidCertificates:            false # Bypass server certificate validation (use with caution)
+                                tlsAllowInvalidHostnames:               false # Bypass server hostname validation (use with caution)
+                                tlsDisableCertificateRevocationCheck:   false # Disable CRL checks
+                                tlsDisableOCSPEndpointCheck:            false # Disable OCSP checks
+                                tlsInsecure:                            false # Allow invalid/no server cert (use with extreme caution)
```
jmikola: Some of these are client-level TLS options. KMS providers only support four options.

config/schema/mongodb-1.0.xsd
GromNaN: Added `keyVaultClient` to the XSD, with a test.

config/schema/mongodb-1.0.xsd
GromNaN: Updated the XSD, 1 `kmsProvider` is required.
https://github.com/doctrine/DoctrineMongoDBBundle/pull/910

config/schema/mongodb-1.0.xsd
GromNaN: Removed.

config/schema/mongodb-1.0.xsd
GromNaN: Added

docs/encryption.rst
GromNaN: Fixed in the ODM: https://github.com/doctrine/mongodb-odm/pull/2801

src/DependencyInjection/Configuration.php
GromNaN: Added in https://github.com/doctrine/DoctrineMongoDBBundle/pull/910

src/DependencyInjection/Configuration.php
GromNaN: Removed in https://github.com/doctrine/DoctrineMongoDBBundle/pull/910

docs/config.rst
GromNaN: I removed `tlsAllowInvalidCertificates`, `tlsAllowInvalidHostnames`, `tlsDisableCertificateRevocationCheck`, and `tlsInsecure` from this list. Taking the list from the [Manager constructor doc](https://www.php.net/manual/en/mongodb-driver-manager.construct.php#mongodb-driver-manager.construct-autoencryption). I checked the list is correct in the `Configuration` class, the tests and the XSD.

docs/encryption.rst
GromNaN: Updated in https://github.com/doctrine/DoctrineMongoDBBundle/pull/912

docs/encryption.rst
GromNaN: Updated in https://github.com/doctrine/DoctrineMongoDBBundle/pull/912

docs/encryption.rst
GromNaN: I removed this. Now `fields` is required in the configuration to match the encrypted fields map format and ensure future compatibility. https://github.com/doctrine/DoctrineMongoDBBundle/pull/912

src/DependencyInjection/Configuration.php
GromNaN: Comment was added just before this line.
```php
// Attribute already present for another KMS type
```

src/DependencyInjection/Configuration.php
GromNaN: That's a bit complex to handle because in the PHP configuration we can have a BSON object (`Binary`) or a JSON representation (`{ $binary: { base64: GH25/XvYSaCgTUQLAo1hQw==, subType: '04' } }`). Doing a deep detection involves a lot of callbacks in the bundle configuration.

src/DependencyInjection/Configuration.php
GromNaN: Exact, the master key is only necessary for creating an encrypted collection. I created a follow-up ticket: [PHPORM-379](https://jira.mongodb.org/browse/PHPORM-379)

src/DependencyInjection/Configuration.php
GromNaN: Added configuration in https://github.com/doctrine/DoctrineMongoDBBundle/pull/913

src/DependencyInjection/Configuration.php
GromNaN: It's already the first option in this list.

src/DependencyInjection/Configuration.php
```diff
@@ -432,7 +433,20 @@ private function addConnectionsSection(ArrayNodeDefinition $rootNode): void
...
                                     ->arrayNode('extraOptions')
-                                        ->prototype('variable')->end()
+                                        ->children()
+                                            ->scalarNode('mongocryptdURI')->end()
+                                            ->booleanNode('mongocryptdBypassSpawn')->end()
+                                            ->scalarNode('mongocryptdSpawnPath')->end()
+                                            ->arrayNode('mongocryptdSpawnArgs')
+                                                ->beforeNormalization()
+                                                    ->ifString()
+                                                    ->then(static fn ($v) => explode(' ', $v))
```
jmikola: If `mongocryptdSpawnArgs` is received as a string in the configuration and blindly split on whitespace, I think we could have edge cases where arguments may be passed to `mongocryptd` incorrectly.

I'll note that the [driver option docs](https://www.php.net/manual/en/mongodb-driver-manager.construct.php) always expect an array of strings, which is what you're normalizing to here.

The BSON array is processed in [`_spawn_mongocryptd()`](https://github.com/mongodb/mongo-c-driver/blob/2.1.0/src/libmongoc/src/mongoc/mongoc-client-side-encryption.c#L1171). Ultimately, the args are received in [`execvp(3)`](https://linux.die.net/man/3/execvp) as a `char**`.

Since _some_ `mongocryptd` args may legitimately contain spaces, I'd worry that allowing the `mongocryptdSpawnArgs` config value to be received as a string could lead to unexpected behavior. Worse, it's possible this could go unnoticed based on how what parsing logic is triggered in `mongocryptd` (e.g. one option might happily accept the unexpected value).

<details>
<summary>See: `mongocryptd -h` output</summary>

````txt
mongocryptd -h
Options:

General options:
  -h [ --help ]                       Show this usage information
  --version                           Show version information
  -f [ --config ] arg                 Configuration file specifying additional 
                                      options
  --configExpand arg                  Process expansion directives in config 
                                      file (none, exec, rest)
  --port arg                          Specify port number - 27020 by default
  --ipv6                              Enable IPv6 support (disabled by default)
  --listenBacklog arg (=128)          Set socket listen backlog size
  --maxConns arg (=1000000)           Max number of simultaneous connections
  --pidfilepath arg                   Full path to pidfile (if not set, no 
                                      pidfile is created)
  --timeZoneInfo arg                  Full path to time zone info directory, 
                                      e.g. /usr/share/zoneinfo
  --nounixsocket                      Disable listening on unix sockets
  --unixSocketPrefix arg              Alternative directory for UNIX domain 
                                      sockets (defaults to /tmp)
  --filePermissions arg               Permissions to set on UNIX domain socket 
                                      file - 0700 by default
  --fork                              Fork server process
  -v [ --verbose ] [=arg(=v)]         Be more verbose (include multiple times 
                                      for more verbosity e.g. -vvvvv)
  --quiet                             Quieter output
  --logpath arg                       Log file to send write to instead of 
                                      stdout - has to be a file, not directory
  --syslog                            Log to system's syslog facility instead 
                                      of file or stdout
  --syslogFacility arg                syslog facility used for mongodb syslog 
                                      message
  --logappend                         Append to logpath instead of over-writing
  --logRotate arg                     Set the log rotation behavior 
                                      (rename|reopen)
  --timeStampFormat arg               Desired format for timestamps in log 
                                      messages. One of iso8601-utc or 
                                      iso8601-local
  --setParameter arg                  Set a configurable parameter

AWS IAM Options:
  --awsIamSessionToken arg            AWS Session Token for temporary 
                                      credentials

MongoCryptD Options:
  --idleShutdownTimeoutSecs arg (=0)  Number of seconds since last received 
                                      request until server shuts down
````txt
</details>

If you decide to accept a string here, I'd probably suggest removing the `explode()` behavior and simply convert the value into a single-element array.

File: tests/DependencyInjection/ConfigurationTest.php [hunk 30]
```diff
@@ -263,7 +263,7 @@ public function testFullConfiguration(array $config): void
                             'mongocryptdURI' => 'mongodb://localhost:27020',
                             'mongocryptdBypassSpawn' => true,
                             'mongocryptdSpawnPath' => '%kernel.project_dir%/bin/mongocryptd',
-                            'mongocryptdSpawnArgs' => '--pidfilepath=%kernel.project_dir%/var/mongocryptd.pid --idleShutdownTimeoutSecs=60',
+                            'mongocryptdSpawnArgs' => ['--pidfilepath=%kernel.project_dir%/var/mongocryptd.pid', '--idleShutdownTimeoutSecs=60'],
```
jmikola, line 266: To my earlier comment, consider if `%kernel.project_dir%` for some reason contained unescaped whitespace. It then might be split into two separate spawn args.

Note: I'm rusty on Symfony's DI container, so I'm not sure if service parameters would be evaluated before or after the bundle configuration's normalization routine.

GromNaN: > To my earlier comment, consider if `%kernel.project_dir%` for some reason contained unescaped whitespace. It then might be split into two separate spawn args.

The `explode` is done very early, when the configuration is read. And the parameter is replaced later when the DI container is dumped. There is no issue is that case.

> Note: I'm rusty on Symfony's DI container, so I'm not sure if service parameters would be evaluated before or after the bundle configuration's normalization routine.

The DI search for parameters in nested arrays. https://github.com/symfony/symfony/blob/8187625c2f5b05fbcc23d74ed66ddc908442e85d/src/Symfony/Component/DependencyInjection/ContainerBuilder.php#L1548-L1549

src/DependencyInjection/Configuration.php
GromNaN: Related to https://github.com/doctrine/DoctrineMongoDBBundle/pull/897#discussion_r2290252539, using a parameter would solve the edge cases.
But I admit there is a bug.

GromNaN: I removed this normalization and modified the XSD so that extra options are XML elements instead of attributes and multiple `<doctrine:mongocryptdSpawnArgs>` can be set.

composer.json
GromNaN:
```suggestion
        "doctrine/mongodb-odm": "^2.6",
```

src/DataCollector/ConnectionDiagnostic.php
jmikola: @jerome: Is this still outstanding? (allowing LB and only prohibiting a standalone)

src/DependencyInjection/Configuration.php
jmikola: Per our video chat, you explained that you intentionally create a non-scalar PHP value here with Extended JSON represented as PHP arrays. Later, in the extension class, you use `json_encode()` and `BsonDocument::fromJSON()` to convert this back into the expected BSON values.

And this is necessary because Symfony's DI container is not able to handle PHP objects (for dumping).

tests/DependencyInjection/Fixtures/config/xml/full.xml
```diff
@@ -67,6 +67,133 @@
...
+                                    "path": "clientCards",
+                                    "bsonType": "array"
+                                }
+                            ]
+                        }
+                    }
+                    ]]>
+                </doctrine:encryptedFieldsMap>
+                <doctrine:extraOptions>
+                    <doctrine:mongocryptdURI>mongodb://localhost:27020</doctrine:mongocryptdURI>
```
jmikola: I think there's a case for making all but `mongocryptdSpawnArgs` attributes. Nested elements should only be needed when multiple values are accepted.

See [`search-index-field`](https://github.com/doctrine/mongodb-odm/blob/2.11.x/doctrine-mongo-mapping.xsd) in the ODM's mapping XSD for prior art.

src/DataCollector/ConnectionDiagnostic.php
```diff
@@ -0,0 +1,96 @@
...
+    public function getServerInfo(): array
+    {
+        $server    = $this->client->getManager()->selectServer(new ReadPreference(ReadPreference::PRIMARY_PREFERRED));
+        $buildInfo = $server->executeCommand('admin', new Command(['buildInfo' => 1]))->toArray()[0] ?? null;
+
+        $version = $buildInfo->version ?? null;
+
+        return [
+            'topologyName' => $this->getTopologyType($server),
+            'topologySupported' => in_array($server->getType(), self::SUPPORTED_SERVER_TYPES),
```
jmikola: Related to my other comment, I think this changes to `!= standalone`. And then `SUPPORTED_SERVER_TYPES` shouldn't be needed.

src/DataCollector/ConnectionDiagnostic.php
jmikola: @GromNaN: This, too.

src/DependencyInjection/Configuration.php
GromNaN: I added a comment.

docs/config.rst
```diff
@@ -825,8 +852,32 @@ Full Default Configuration
 
             $config->connection('id')
                 ->server('mongodb://localhost')
-                ->driverOptions([
-                    'context' => null, // stream context to use for connection
+                ->driverOptions(['context' => null]), // stream context to use for connection
```
jmikola: Note that `ext-mongodb` 2.0 removed the `context` driver option (see: [`Manager::__construct()`](https://www.php.net/manual/en/mongodb-driver-manager.construct.php)). Feel free to remove this from the docs in this PR (both here and above), or create another PR/issue to do so later.

docs/encryption.rst
```diff
@@ -0,0 +1,307 @@
...
+--------------------
+
+The encrypted fields are set to the collection when you create it, and the MongoDB
+client will query the server for the collection schema before performing any
+operations. For additional security, you can also specify the encrypted fields
+in the connection configuration, which allows the client to use local rules
+instead of downloading the remote schema from the server, that could potentially
+be tampered with if an attacker compromises the server.
+
+The Encrypted Fields Maps is a list of all encrypted fields associated with all
```
jmikola:
```suggestion
The Encrypted Fields Map is a list of all encrypted fields associated with all
```

docs/encryption.rst
```diff
@@ -0,0 +1,307 @@
...
+
+            return static function (DoctrineMongodbConfig $config): void {
+                $config->connection('default')
+                    ->autoEncryption([
+                        'encryptedFieldsMap' => [
+                            'mydatabase.mycollection' => [
+                                'fields' => [
+                                    [
+                                        'path' => 'sensitive_field',
+                                        'keyId' => ['$binary' => ['base64' => '2CSosXLSTEKaYphcSnUuCw==', 'subType' => '04' ] ],
```
jmikola, line 168: IIRC, you said that a Binary object _cannot_ be used here. Do you want to note this, as it may be unintuitive for users configuring the bundle in PHP. If anything, you may want to point them to the [Extended JSON spec](https://github.com/mongodb/specifications/blob/master/source/extended-json/extended-json.md).

docs/config.rst
```diff
@@ -825,8 +852,32 @@ Full Default Configuration
...
+                    'bypassQueryAnalysis' => false, // Disables automatic analysis of read and write operations for encryption
+                    'tlsOptions' => [        // TLS options for the Key Vault client (if keyVaultClient is not specified)
+                        // 'tlsCAFile' => null,  // Path to CA file, e.g., /path/to/key-vault-ca.pem
+                        // 'tlsCertificateKeyFile' => null,  // Path to client cert/key file, e.g., /path/to/key-vault-client.pem
+                        // 'tlsCertificateKeyFilePassword' => null,  // Password for client cert/key file
+                        // 'tlsAllowInvalidCertificates' => false, // Bypass server certificate validation (use with caution)
+                        // 'tlsAllowInvalidHostnames' => false, // Bypass server hostname validation (use with caution)
+                        // 'tlsDisableCertificateRevocation' => false, // Disable CRL checks
+                        // 'tlsDisableOCSPEndpointCheck' => false, // Disable OCSP checks
+                        // 'tlsInsecure' => false, // Allow invalid/no server cert (use with extreme caution)
```
jmikola: I think you forgot to clean up this list after restricting `tlsOptions` to the four supported options.

docs/encryption.rst
```diff
@@ -0,0 +1,307 @@
...
+                                tlsCertificateKeyFilePassword: "keyvaultclientpassword"
+                                tlsDisableOCSPEndpointCheck: false
+
+    .. group-tab:: XML
+
+        .. code-block:: xml
+
+            <doctrine:connection>
+                <doctrine:autoEncryption>
+                    <doctrine:tlsOptions>
```
jmikola: This doesn't match the XSD, where you changed the TLS options to XML attributes.

Might as well confirm that all XML examples in the docs still conform to the XSD, as I believe you moved other things to attributes as well.

docs/encryption.rst
```diff
@@ -0,0 +1,307 @@
...
+                        'tlsOptions' => [
+                            'tlsCAFile' => '/path/to/key-vault-ca.pem',
+                            'tlsCertificateKeyFile' => '/path/to/key-vault-client.pem',
+                            'tlsCertificateKeyFilePassword' => 'keyvaultclientpassword',
+                            'tlsDisableOCSPEndpointCheck' => false,
+                        ],
+                    ]);
+            };
+
+Context Service for SSL
```
jmikola: This seems unrelated to Encryption. Was it copied from another docs file? If so, you can remove it from _both_ places since `context` was dropped in `ext-mongodb` 2.0 (and deprecated long before then).

tests/DependencyInjection/Fixtures/config/yml/full.yml
```diff
@@ -55,6 +55,90 @@ doctrine_mongodb:
...
                 context:          conn1_context_service
+            autoEncryption:
+                kmsProvider:
+                    type: 'aws'
+                    accessKeyId: 'MONGODB_AWS_ACCESS_KEY_ID'
+                    secretAccessKey: 'MONGODB_AWS_SECRET_ACCESS_KEY'
+                    sessionToken: 'MONGODB_AWS_SESSION_TOKEN'
+                masterKey:
+                    key: 'MONGODB_AWS_MASTER_KEY'
+                keyVaultNamespace: 'encryption.__keyVault'
```
jmikola, line 66: Noted that you're using the namespace from the server manual here. 👍

config/schema/mongodb-1.0.xsd
```diff
@@ -119,12 +123,90 @@
...
+    <!-- Attribute already present for another KMS type -->
+    <!-- <xsd:attribute name="keyName" type="xsd:string" /> -->
+    <!-- <xsd:attribute name="keyVersion" type="xsd:string" /> -->
+    <!-- KMIP -->
+    <!-- <xsd:attribute name="endpoint" type="xsd:string" /> -->
+    <!-- Local -->
+    <xsd:attribute name="key" type="xsd:string" />
+  </xsd:complexType>
+
+  <xsd:complexType name="master-key">
```
jmikola, line 170: Does this warrant its own complex type? Why not just allow the master key to be specified as an attribute on the parent `masterKey` XML element?

That would be consistent with the other string options at that level, such as `keyVaultClient` and `keyVaultNamespace`.

src/Command/DumpEncryptedFieldsMapCommand.php
```diff
@@ -0,0 +1,113 @@
...
+        parent::__construct();
+    }
+
+    protected function configure(): void
+    {
+        $this->addOption(
+            'format',
+            'f',
+            InputOption::VALUE_REQUIRED,
+            'The output format for the encrypted fields map (yaml, php)',
```
jmikola: Is "json" intentionally omitted from the parenthetical here?

src/DependencyInjection/Configuration.php
```diff
@@ -339,6 +342,147 @@ private function addConnectionsSection(ArrayNodeDefinition $rootNode): void
...
+                                    ->arrayNode('masterKey')
+                                        ->prototype('variable')->end()
+                                    ->end()
+                                    ->arrayNode('kmsProvider')
+                                        ->isRequired()
+                                        ->children()
+                                            ->scalarNode('type')
+                                                ->isRequired()
+                                                ->validate()
+                                                    ->ifTrue(static fn ($v) => ! in_array($v, ['aws', 'azure', 'gcp', 'kmip', 'local'], true))
```
jmikola, line 364: Noted that [named KMS providers](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.md#kmsproviders) are also not supported. I think we may have discussed that when I asked whether _multiple_ providers could be specified.

src/DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -382,9 +400,10 @@ protected function loadDocumentManager(array $documentManager, string|null $defa
      * @param array            $config    An array of connections configurations
      * @param ContainerBuilder $container A ContainerBuilder instance
      */
-    protected function loadConnections(array $connections, ContainerBuilder $container): void
+    protected function loadConnections(array $connections, ContainerBuilder $container, array $config): void
     {
-        $cons = [];
+        $cons        = [];
+        $diagnostics = [];
```
jmikola: Is this used?

src/DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -484,6 +514,54 @@ private function normalizeDriverOptions(array $connection): array
...
+
+        $provider     = $autoEncryption['kmsProvider']['type'];
+        $providerOpts = array_diff_key($autoEncryption['kmsProvider'], ['type' => true]);
+        // To use "Automatic Credentials", the provider options must be an empty document.
+        // Fix the empty array to an empty stdClass object, as the driver expects it.
+        if ($providerOpts === []) {
+            $providerOpts = new Definition('stdClass');
+        }
+
+        $autoEncryption['kmsProviders'] = [$provider => $providerOpts];
```
jmikola, line 542: Why does this assign a `kmsProviders` key, which is later removed by the `array_diff_key()` in `loadDocumentManager()` (when calling `normalizeAutoEncryption()`)?

In the calling context, `$authEncryption['kmsProvider']` still gets used to queue up a call to `setKmsProvider()`, but it doesn't look like any of the normalization treatment above will affect that.

Please let me know if I'm overlooking something.

tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
```diff
@@ -479,4 +485,182 @@ public function testUseTransactionalFlush(): void
...
+        $dummyServiceId = 'my_key_vault_client_service';
+        $container->setDefinition($dummyServiceId, new Definition(Client::class));
+
+        $config = [
+            'connections' => [
+                'default' => [
+                    'autoEncryption' => [
+                        'keyVaultNamespace' => 'db.vault',
+                        'keyVaultClient' => $dummyServiceId,
+                        'kmsProvider' => ['type' => 'local', 'key' => 'cGFzc3dvcmQ='],
```
jmikola: `cGFzc3dvcmQ=` is "password" after base64 decoding. Should this actually be a 96-byte string? I suppose it doesn't really matter since you're not validating that in the configuration. And I assume that's intentional and you're relying on the server to complain (not that it matters for this unit test).

tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
```diff
@@ -479,4 +485,182 @@ public function testUseTransactionalFlush(): void
...
+            array_diff_key($odmConfiguration->getDriverOptions()['autoEncryption'], ['keyVaultClient' => false]),
+        );
+    }
+
+    public function testAutoEncryptionWithComplexKmsAndSchemaMap(): void
+    {
+        $container = $this->buildMinimalContainer();
+        $loader    = new DoctrineMongoDBExtension();
+
+        $schemaMap = [
```
jmikola, line 549: The `schemaMap` option isn't presented in any ODM docs AFAIK. Is there a reason to test for this here? I assume you're just verifying that the option itself makes it through untouched?

tests/DependencyInjection/Fixtures/config/xml/full.xml
```diff
@@ -7,8 +7,8 @@
                         http://symfony.com/schema/dic/doctrine/odm/mongodb http://symfony.com/schema/dic/doctrine/odm/mongodb/mongodb-1.0.xsd">
 
     <doctrine:config
-        auto-generate-hydrator-classes="true"
-        auto-generate-proxy-classes="true"
+        auto-generate-hydrator-classes="1"
+        auto-generate-proxy-classes="2"
```
jmikola, line 11: Why did you make these replacements? In `full.yml` below you have:

```yaml
auto_generate_proxy_classes: 2
auto_generate_hydrator_classes: true
auto_generate_persistent_collection_classes: 3
```

config/schema/mongodb-1.0.xsd
GromNaN: The master key is an array, not a string. 
https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBDatabase-createEncryptedCollection/

tests/DependencyInjection/Fixtures/config/xml/full.xml
GromNaN: Because the xsd expects `int`
https://github.com/doctrine/DoctrineMongoDBBundle/blob/fd7d666762bcb935dfc218973db050b38802f645/config/schema/mongodb-1.0.xsd#L17-L18

This was missing from https://github.com/doctrine/DoctrineMongoDBBundle/pull/337

tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
GromNaN: Excactly, I just not tried to work with the schemaMap.

src/DependencyInjection/DoctrineMongoDBExtension.php
GromNaN: ~Good catch, we don't need to set the `kmsProviders` here.~

The method `normalizeAutoEncryption` is also used to generate `driverOptions` that are injected into the client constructor. We cannot use the `Configuration` class to get the driver options because the `Client` connection is independent from the document manager and its configuration.

src/DependencyInjection/Configuration.php
GromNaN: The named providers are only useful when you need multiple KMS providers of the same type. But since we support only 1 KMS provider, we don't need to support names.

docs/encryption.rst
GromNaN: Adding this note:
```
// Extended JSON representation of a BSON binary type
// The MongoDB\BSON\Binary class cannot be used here
```

docs/config.rst
GromNaN: I remove this from the config example. I marked this config as deprecated in https://github.com/doctrine/DoctrineMongoDBBundle/pull/893

src/DependencyInjection/DoctrineMongoDBExtension.php
jmikola: In that case, shouldn't the `normalizeDriverOptions` calling context use its own `array_diff_key()` to unset the `kmsProvider` key specific to ODM? Otherwise, I think you're passing an unrecognized option to PHPLIB and, ultimately, PHPC. It'd be silently ignored now but could become a possible low-level log message down the line.

config/schema/mongodb-1.0.xsd
jmikola: Based on this XSD,

There are various struct formats (by provider) for the `masterKey` option, as documented in [ClientEncryption::createDataKey()](https://www.php.net/mongodb-driver-clientencryption.createdatakey).

The XSD defines the `master-key` complex type as:

```
<xsd:attribute name="key" type="xsd:string" use="required" />
```

Unless I'm mistaken (and my XSD is rusty), that's going to allow a single `key` string option to be specified within `masterKey`. And that's just one of several sub-options for the "aws" provider.

Should this be changed to allow any number of `<string-option>` and `<bool-option>` values? (names there are for example purposes, but maybe there is a general need for since this is a recurring problem for complex configs in XML).

tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
jmikola: @GromNaN: Feel free to resolve if my assumption above is correct.

src/Command/EncryptionDumpFieldsMapCommand.php
```diff
@@ -76,9 +76,24 @@ protected function execute(InputInterface $input, OutputInterface $output): int
                 continue;
             }
 
+            // The min/max query options must have the same type as the field.
+            // But the PHP driver always convert to "int" or "float" when the value fit in the range
             foreach ($encryptedFieldsMap as $ns => $encryptedFields) {
+                $fields = json_decode(PackedArray::fromPHP($encryptedFields['fields'])->toCanonicalExtendedJSON(), true);
```
jmikola: Would it be prudent to specify [`JSON_THROW_ON_ERROR`](https://www.php.net/manual/en/json.constants.php#constant.json-throw-on-error) here. It looks like that is commonly done [in Symfony](https://github.com/search?q=repo%3Asymfony%2Fsymfony%20JSON_THROW_ON_ERROR&type=code).

jmikola (approved):
A few questions, but you can certainly address those on your own (or rule out my concerns). I do think some XSD changes might be required for `masterKey`, though.

Other than that, note that there are still two unchecked TODOs in the PR description. I noted #902 is deferred. The other to bump `composer.json` might already have been done, and just needs checking.

tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
GromNaN: Yes, that's just a random string (actually provided by copilot).

src/Command/EncryptionDumpFieldsMapCommand.php
GromNaN: The JSON comes from `toCanonicalExtendedJSON`; it will be always valid.

jmikola: Famous last words :)

But yes, I'm OK with leaving this as-is.

tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
jmikola: > (actually provided by copilot)

In that case I'm going to assume it was actually leaked credentials from another repository that snuck into the Copilot training data. 😆

src/DependencyInjection/DoctrineMongoDBExtension.php
GromNaN: I had the idea that we could use the `ODMConfiguration` class at build time to generate the `driverOptions`. But that would not work very well with environment variables. And would increase the required version of `doctrine/mongodb-odm`.

PR #898, opened 2025-06-18
[Encryption] Improve diagnostic command

GromNaN:
From https://github.com/GromNaN/DoctrineMongoDBBundle/pull/1

> This improves the output of the diagnostics command and adds a command that prints the encrypted fields map for use in configuration.

Output: 

```
❯ bin/console doctrine:mongodb:connection:diagnostic 

MongoDB Encryption Diagnostics
==============================

 PHP Environment
 * MongoDB extension loaded: Yes
 * MongoDB extension version: 1.21.1
 * MongoDB extension supports libmongocrypt: Yes
 * MongoDB library version: 2.1.0

 mongocryptd information
 * mongocryptd path: /Users/jerome/.local/bin/mongocryptd
 * mongocryptd version: mongocrypt version v8.0.8

Connection: default
-------------------

 Server Information
 * Server Version: 8.0.8
 * Topology: Replica Set

 Auto Encryption Configuration
 * Auto Encryption Enabled: Yes
 * Key Vault Namespace: symfony.datakeys
 * Key Count: 0

                                                                                                         
 [OK] System looks ok for encryption support.                                                            
                                                                                                         
```

In case of incorrect setup:

```
MongoDB Encryption Diagnostics
==============================

 PHP Environment
 * MongoDB extension loaded: Yes
 * MongoDB extension version: 2.1.0
 * MongoDB extension supports libmongocrypt: Yes
 * MongoDB library version: 2.1.0

 mongocryptd information
 * mongocryptd: not found

Connection: default
-------------------

 Server Information
 * Server Version: 8.0.8
 * Topology: Standalone

                                                                                                         
 [WARNING] This topology does not support encryption.                                                    
                                                                                                         

 Auto Encryption Configuration

                                                                                                         
 [ERROR] Failed to retrieve auto encryption information: mongocryptd error: No suitable servers found:   
         `serverSelectionTimeoutMS` expired: [connection refused calling hello on 'localhost:27020'].    
         Topology type: Single:                                                                          
                                                                                                         

Connection: other
-----------------

 Server Information
 * Server Version: 8.0.8
 * Topology: Standalone

                                                                                                         
 [WARNING] This topology does not support encryption.                                                    
                                                                                                         

 Auto Encryption Configuration
 Auto encryption is not enabled for this connection.

                                                                                                         
 [WARNING] Not all requirements for encryption support are met. Please check the diagnostics above.      
```

### DIFF
--- a/docs/encryption.rst
+++ b/docs/encryption.rst
@@ -76,10 +76,23 @@ Example of configuration for AWS
                         key: "arn:aws:kms:eu-west-1:123456789012:key/abcd1234-12ab-34cd-56ef-1234567890ab"
 
 
-Queryable Encryption (QE)
--------------------------
+Encrypted Fields Map
+--------------------
 
-Queryable Encryption (QE) allows you to run queries on encrypted fields. To use QE, you may need to provide an ``encryptedFieldsMap`` or use a schema map, depending on your driver and use case.
+You can configure which fields are encrypted in each collection by specifying the
+``autoEncryption.encryptedFieldsMap`` option in the connection configuration.
+This setting is **recommended** for improved security and performance.
+
+- If the connection ``encryptedFieldsMap`` object contains a key for the specified
+  collection, the client uses that object to perform automatic Queryable Encryption,
+  rather than using the remote schema. At minimum, the local rules must encrypt
+  all fields that the remote schema does.
+
+- If the connection ``encryptedFieldsMap`` object doesn't contain a key for the
+  specified collection, the client downloads the server-side remote schema for
+  the collection and uses it instead.
+
+For more details, see the official MongoDB documentation: `Encrypted Fields and Enabled Queries <https://www.mongodb.com/docs/manual/core/queryable-encryption/fundamentals/encrypt-and-query/>`_.
 
 .. tabs::
 
@@ -129,6 +142,51 @@ Queryable Encryption (QE) allows you to run queries on encrypted fields. To use 
                                     ],
                                 ],
                             ],
+                        ],
+                    ]);
+            };
+
+Automatic Encryption Shared Library
+-----------------------------------
+
+To use automatic encryption, the MongoDB PHP driver requires the `Automatic Encryption Shared Library`_.
+
+If the driver is not able to find the library, you can specify its path using the ``cryptSharedLibPath`` extra option in your connection configuration.
+
+.. tabs::
+
+    .. group-tab:: YAML
+
+        .. code-block:: yaml
+
+            doctrine_mongodb:
+                connections:
+                    default:
+                        autoEncryption:
+                            extraOptions:
+                                cryptSharedLibPath: '%kernel.project_dir%/bin/mongo_crypt_v1.so'
+
+    .. group-tab:: XML
+
+        .. code-block:: xml
+
+            <doctrine:connection>
+                <doctrine:autoEncryption>
+                    <doctrine:extraOptions cryptSharedLibPath="%kernel.project_dir%/bin/mongo_crypt_v1.so" />
+                </doctrine:autoEncryption>
+            </doctrine:connection>
+
+    .. group-tab:: PHP
+
+        .. code-block:: php
+
+            use Symfony\Config\DoctrineMongodbConfig;
+
+            return static function (DoctrineMongodbConfig $config): void {
+                $config->connection('default')
+                    ->autoEncryption([
+                        'extraOptions' => [
+                            'cryptSharedLibPath' => '%kernel.project_dir%/bin/mongo_crypt_v1.so',
                         ],
                     ]);
             };
@@ -221,3 +279,5 @@ Further Reading
 - `MongoDB CSFLE documentation <https://www.mongodb.com/docs/manual/core/csfle/>`_
 - `MongoDB PHP driver Manager::__construct <https://www.php.net/manual/en/mongodb-driver-manager.construct.php>`_
 - :doc:`config`
+
+.. _`Automatic Encryption Shared Library`: https://www.mongodb.com/docs/manual/core/queryable-encryption/install-library/
--- a/src/Command/ConnectionDiagnosticCommand.php
+++ b/src/Command/ConnectionDiagnosticCommand.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace Doctrine\Bundle\MongoDBBundle\Command;
 
 use Doctrine\Bundle\MongoDBBundle\DataCollector\ConnectionDiagnostic;
+use Doctrine\Bundle\MongoDBBundle\DataCollector\EncryptionDiagnostic;
+use MongoDB\Driver\Exception\RuntimeException;
 use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
@@ -12,7 +14,6 @@ use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Contracts\Service\ServiceProviderInterface;
-use Throwable;
 
 use function array_diff;
 use function array_keys;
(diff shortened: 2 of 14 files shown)


src/Command/ConnectionDiagnosticCommand.php
```diff
@@ -54,59 +57,18 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
-                        'Key Vault Namespace: ' . $autoEncryptionInfo['keyVaultNamespace'],
-                        'Key Count: ' . $autoEncryptionInfo['keyCount'],
-                    ]);
-                } else {
-                    $io->text('No auto encryption configuration found for this connection.');
-                }
-            } catch (Throwable $exception) {
-                $io->error('Could not retrieve auto encryption info: ' . $exception->getMessage());
-            }
+            $configOk   = $this->printAndCheckConnectionDiagnostic($name, $diagnostic, $io) && $configOk;
```
GromNaN, line 65: When multiple connections are configured, but they don't have all auto encryption enabled, then the conclusion says there is an issue. We should return "OK" when auto encryption is not configured on a connection?

alcaeus: Ah, that's a good point. Indeed, if encryption wasn't configured on a connection, we should print an info line and mark it as "ok".

src/DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -517,6 +503,16 @@ private function normalizeDriverOptions(array $connection, array $config): array
...
+
+            if (isset($driverOptions['autoEncryption']['extraOptions']['cryptSharedLibPath'])) {
+                $fs                 = new Filesystem();
+                $cryptSharedLibPath = $driverOptions['autoEncryption']['extraOptions']['cryptSharedLibPath'];
+
+                // If it's not an absolute path, resolve it relative to the project root
+                if (! $fs->isAbsolutePath($cryptSharedLibPath)) {
+                    $driverOptions['autoEncryption']['extraOptions']['cryptSharedLibPath'] = realpath($cryptSharedLibPath);
+                }
+            }
```
GromNaN: @alcaeus I'm reverting this part. In Symfony we never convert relative paths to absolute.

I'll add documentation using `%kernel.project_dir%` instead.

src/Command/ConnectionDiagnosticCommand.php
```diff
@@ -27,8 +28,10 @@
 final class ConnectionDiagnosticCommand extends Command
 {
     /** @param ServiceProviderInterface<ConnectionDiagnostic> $diagnostics */
-    public function __construct(private readonly ServiceProviderInterface $diagnostics)
-    {
+    public function __construct(
+        private readonly ServiceProviderInterface $diagnostics,
+        private readonly EncryptionDiagnostic $encryptionDiagnostic = new EncryptionDiagnostic(),
```
GromNaN, line 33: We don't declare a service for `EncryptionDiagnostic`. Only leverage the default value from the constructor.

src/DependencyInjection/DoctrineMongoDBExtension.php
alcaeus: Sounds good - I wasn't quite sure if the kernel parameter was resolved properly. If that works, I prefer that over changing the path in the extension.

docs/encryption.rst
```diff
@@ -133,6 +147,51 @@ Queryable Encryption (QE) allows you to run queries on encrypted fields. To use
...
+    .. group-tab:: YAML
+
+        .. code-block:: yaml
+
+            doctrine_mongodb:
+                connections:
+                    default:
+                        autoEncryption:
+                            extraOptions:
+                                cryptSharedLibPath: '%kernel.project_dir%/bin/libmongocrypt.so'
```
alcaeus: Where did you get the name from? I would like to avoid using `libmongocrypt`, as that's the name of the library that the drivers use internally and is separate from crypt_shared, which is used by libmongocrypt. The file I downloaded for my Mac is named `mongo_crypt_v1.dylib`, for Linux it's `mongo_crypt_v1.so`. We should use one of those names for clarity.

GromNaN: Fixed. I think copilot generated the initial name in the tests, and I copied here. Using the actual file name is a lot better. Thanks!

paulinevos:
There's a lot of commits in this. Does it need to be rebased?

GromNaN:
> There's a lot of commits in this. Does it need to be rebased?

It will be squashed on merge. All the commits represent the iterations on this PR.

src/Command/ConnectionDiagnosticCommand.php
```diff
@@ -154,11 +155,17 @@ private function printAutoEncryptionConfiguration(SymfonyStyle $io, ConnectionDi
...
-        ]);
+        try {
+            $autoEncryptionInfo = $diagnostic->getAutoEncryptionInfo();
+
+            $io->listing([
+                'Auto Encryption Enabled: ' . ($autoEncryptionInfo['autoEncryptionEnabled'] ? 'Yes' : 'No'),
+                'Key Vault Namespace: ' . $autoEncryptionInfo['keyVaultNamespace'],
+                'Key Count: ' . $autoEncryptionInfo['keyCount'],
+            ]);
+        } catch (RuntimeException $e) {
```
paulinevos, line 173: Why not catch something broader? I'm guessing there could be other exception types that we'd want to catch here?

docs/encryption.rst
```diff
@@ -76,10 +76,24 @@ Example of configuration for AWS
...
 
-Queryable Encryption (QE)
--------------------------
+Encrypted Fields Map
+--------------------
 
-Queryable Encryption (QE) allows you to run queries on encrypted fields. To use QE, you may need to provide an ``encryptedFieldsMap`` or use a schema map, depending on your driver and use case.
+You can configure which fields are encrypted in each collection by specifying the
+``autoEncryption.encryptedFieldsMap`` option in the connection configuration.
+This setting **recommended** for improved security and performance.
```
paulinevos: Missing `is` after `this setting`

src/DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -44,7 +44,6 @@
 use Symfony\Component\DependencyInjection\Loader\PhpFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
-use Symfony\Component\Filesystem\Filesystem;
```
paulinevos: I would use fixup to add this commit to either:
- whichever commit introduced the CS errors if that commit is part of this PR, or
- fix it up to the other `fix CS` commit you already have

Not a big deal not but makes reviewing a little easier to go commit by commit

src/Command/ConnectionDiagnosticCommand.php
```diff
@@ -80,16 +80,21 @@ private function getConnectionNames(): array
...
     }
 
+    /** @return bool True if the server is compatible with auto-encryption configuration, false otherwise. */
     private function printAndCheckConnectionDiagnostic(string $name, ConnectionDiagnostic $diagnostic, SymfonyStyle $io): bool
     {
         $io->section(sprintf('Connection: %s', $name));
 
-        $configOk = $this->printAndCheckServerInfo($io, $diagnostic);
-        $this->printAutoEncryptionConfiguration($io, $diagnostic);
+        $enabled = $this->printAutoEncryptionConfiguration($io, $diagnostic);
```
paulinevos: Maybe rename this to `autoEncryptionEnabled`?

paulinevos (changes requested):
some small comments, otherwise seems fine

docs/encryption.rst
GromNaN:
```suggestion
This setting is **recommended** for improved security and performance.
```

src/Command/ConnectionDiagnosticCommand.php
GromNaN: I only want to catch driver errors due to the command that is run. I don't expect other kind of exception to occurs, so if there is an other error I don't want to catch it, but I want the regular error handler to display it (with the backtrace).

src/DependencyInjection/DoctrineMongoDBExtension.php
GromNaN: There was so many iteration, I think reviewing commit-by-commit doesn't really make sense. I squashed.

paulinevos (approved):

PR #905, opened 2025-07-04
[Encryption] Fix format of `encryptedFieldsMaps` in the `autoEncryption` configuration

GromNaN:
In driver options, the `encryptedFieldsMap` option must:
1. contain a `fields` key with the list of fields
2. convert `query.min` and `query.max` to the same BSON type as the field
3. contain the `keyId`. But this value will change every time the database is reset.

This needs to be improved by leveraging `keyAltNames`.


src/Command/DumpEncryptedFieldsMapCommand.php
```diff
@@ -63,24 +61,27 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
 
         foreach ($this->documentManagers as $name => $documentManager) {
-            $generator          = new EncryptedFieldsMapGenerator($documentManager->getMetadataFactory());
-            $encryptedFieldsMap = $generator->getEncryptedFieldsMap();
+            assert($documentManager instanceof DocumentManager);
+
+            $encryptedFieldsMap = [];
+            foreach ($documentManager->getMetadataFactory()->getAllMetadata() as $metadata) {
+                $database               =  $documentManager->getDocumentDatabase($metadata->getName());
+                $collectionInfoIterator = $database->listCollections(['filter' => ['name' => $metadata->getCollection()]]);
```
GromNaN, line 66: This is inspired by the [`get_encrypted_fields_from_server` function](https://github.com/mongodb/mongo-php-library/blob/0058bc008028d91f5f9e032e80daa094b31ef9a9/src/functions.php#L201).

src/DependencyInjection/Configuration.php
```diff
@@ -424,13 +424,16 @@ private function addConnectionsSection(ArrayNodeDefinition $rootNode): void
                                                 ->children()
                                                     ->scalarNode('path')->isRequired()->cannotBeEmpty()->end()
                                                     ->scalarNode('bsonType')->isRequired()->cannotBeEmpty()->end()
+                                                    ->variableNode('keyId')->defaultNull()->end()
```
GromNaN: The configuration looks like this (with binary UUID and BSON values for min and max)

```yaml
                encryptedFieldsMap:
                  symfony.RangeTypes:
                    - keyId: { $binary: { base64: lhZHItpvRkqXevh4Wtqg/g==, subType: '04' } }
                      path: intField
                      bsonType: int
                      queries: { queryType: range, contention: 8, min: 5, max: 10 }
                    - keyId: { $binary: { base64: qd9PEKIPTE2J30ev29lMpQ==, subType: '04' } }
                      path: floatField
                      bsonType: double
                      queries: { queryType: range, contention: 8, min: 5.5, max: 10.5, precision: 1 }
                    - keyId: { $binary: { base64: zVLg8CF4RSSu4xn7x7dOyQ==, subType: '04' } }
                      path: decimalField
                      bsonType: decimal
                      queries: { queryType: range, contention: 8, min: { $numberDecimal: '0.1' }, max: { $numberDecimal: '0.2' }, precision: 2 }
                    - keyId: { $binary: { base64: ySdd8lZ2QBqnwKPJTp/yLA==, subType: '04' } }
                      path: immutableDateField
                      bsonType: date
                      queries: { queryType: range, contention: 8, min: { $date: '2000-01-01T00:00:00Z' }, max: { $date: '2100-01-01T00:00:00Z' } }
                    - keyId: { $binary: { base64: NWKI+DyES/OlNkUbJbWJ9w==, subType: '04' } }
                      path: dateField
                      bsonType: date
                    - keyId: { $binary: { base64: wiiv+0K/QAquyEq3HDxRKw==, subType: '04' } }
                      path: binField
                      bsonType: binData
                    - keyId: { $binary: { base64: 2CSosXLSTEKaYphcSnUuCw==, subType: '04' } }
                      path: timestampField
                      bsonType: timestamp
                    - keyId: { $binary: { base64: h3H6HdG3T5CK+Z2yQ4Ho+Q==, subType: '04' } }
                      path: hashField
                      bsonType: object
                    - keyId: { $binary: { base64: X78UZZ/HTX2wLw4K3uG42w==, subType: '04' } }
                      path: collectionField
                      bsonType: objectId
                    - keyId: { $binary: { base64: LugQL/ZXTJOl856Yacmkwg==, subType: '04' } }
                      path: boolField
                      bsonType: bool
                  symfony.patients:
                    - keyId: { $binary: { base64: GH25/XvYSaCgTUQLAo1hQw==, subType: '04' } }
                      path: pathologies
                      bsonType: array
                    - keyId: { $binary: { base64: krVWyFlNTUOaGFMfk+s7UA==, subType: '04' } }
                      path: patientRecord.billing
                      bsonType: object
                    - keyId: { $binary: { base64: X1ZaSI1GSAKnZ+sPGcmYBA==, subType: '04' } }
                      path: patientRecord.billingAmount
                      bsonType: int
                      queries: { queryType: range, contention: 8, min: 100, max: 2000, sparsity: 1, trimFactor: 4 }
                  symfony.client:
                    - keyId: { $binary: { base64: I0Aw18vnRGWzVS1t3uejpQ==, subType: '04' } }
                      path: name
                      bsonType: string
                    - keyId: { $binary: { base64: XSPRK3vaTLmMZr9IEj/qwQ==, subType: '04' } }
                      path: clientCards
                      bsonType: array
```

src/DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -538,6 +540,13 @@ private function normalizeAutoEncryption(array $autoEncryption, string $defaultD
 
         $autoEncryption['keyVaultNamespace'] ??= $defaultDB . '.datakeys';
 
+        if (isset($autoEncryption['encryptedFieldsMap'])) {
+            foreach ($autoEncryption['encryptedFieldsMap'] as &$value) {
+                //$value = ['fields' => $value];
```
alcaeus: This commented line should be removed.

alcaeus (commented):
I think this change is reasonable. We need to make sure how to properly generate the encryptedFieldsMap for use in the configuration. For a future improvement, we can consider dumping the `encryptedFieldsMap` based on the return value of `createEncryptedCollection` calls in the `schema:create` command.

tests/DependencyInjection/Fixtures/config/xml/full.xml
```diff
@@ -86,20 +86,96 @@
...
-                        <doctrine:field path="patientRecord.billingAmount" bsonType="int">
-                            <doctrine:queries queryType="range" min="100" max="2000" sparsity="1" trimFactor="4" />
-                        </doctrine:field>
-                    </doctrine:encryptedFields>
-                    <doctrine:encryptedFields name="encrypted.users">
-                        <doctrine:field path="email" bsonType="string">
-                            <doctrine:queries queryType="equality" />
-                        </doctrine:field>
-                    </doctrine:encryptedFields>
+                    <![CDATA[
```
GromNaN, line 89: @alcaeus Since the XML format is set to disappear, I have simplified the configuration here. EncryptedFieldsMap is a JSON string that is copied from the dump command in the XML. It is too complicated to convert this to XML tags, and most likely unusable anyway.

alcaeus: That makes sense 👍

docs/encryption.rst
```diff
@@ -135,11 +146,10 @@ For more details, see the official MongoDB documentation: `Encrypted Fields and
...
                         'encryptedFieldsMap' => [
                             'mydatabase.mycollection' => [
-                                'fields' => [
-                                    [
-                                        'path' => 'sensitive_field',
-                                        'bsonType' => 'string',
-                                    ],
+                                [
+                                    'path' => 'sensitive_field',
+                                    'keyId' => ['$binary' => ['base64' => '2CSosXLSTEKaYphcSnUuCw==', 'subType' => '04' ] ],
```
alcaeus, line 151: For PHP, should we suggest people create a `Binary` instance?
```suggestion
                                    'keyId' => new Binary(base64_decode('2CSosXLSTEKaYphcSnUuCw=='), Binary::TYPE_UUID),
```

On second thought after reviewing the rest of this PR, I'm guessing this format is created by the PHP dumper in the command that dumps the encryptedFieldsMap. Feel free to disregard in that case.

src/Command/DumpEncryptedFieldsMapCommand.php
```diff
@@ -63,24 +60,25 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $dumper = new Dumper();
 
         foreach ($this->documentManagers as $name => $documentManager) {
-            $generator          = new EncryptedFieldsMapGenerator($documentManager->getMetadataFactory());
-            $encryptedFieldsMap = $generator->getEncryptedFieldsMap();
+            $encryptedFieldsMap = [];
+            foreach ($documentManager->getMetadataFactory()->getAllMetadata() as $metadata) {
+                $database               =  $documentManager->getDocumentDatabase($metadata->getName());
```
alcaeus:
```suggestion
                $database               = $documentManager->getDocumentDatabase($metadata->getName());
```

src/Command/DumpEncryptedFieldsMapCommand.php
```diff
@@ -63,24 +60,25 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+                $collectionInfoIterator = $database->listCollections(['filter' => ['name' => $metadata->getCollection()]]);
+
+                foreach ($collectionInfoIterator as $collectionInfo) {
+                    if ($collectionInfo['options']['encryptedFields'] ?? null) {
+                        $encryptedFieldsMap[$this->getDocumentNamespace($metadata, $database->getDatabaseName())] = $collectionInfo['options']['encryptedFields'];
```
alcaeus, line 70: Do we want to merge encrypted fields maps if one already exists? This could happen in inheritance scenarios. Not sure if we want to consider the use case and its implications, but it wouldn't be too hard to ensure that all fields are encrypted.

src/Command/DumpEncryptedFieldsMapCommand.php
```diff
@@ -63,24 +60,25 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
-                    fn (string $fqcn): string => $this->getDocumentNamespace(
-                        $documentManager->getClassMetadata($fqcn),
-                        $documentManager->getConfiguration()->getDefaultDB(),
-                    ),
-                    array_keys($encryptedFieldsMap),
-                ),
-                array_values($encryptedFieldsMap),
-            );
+            foreach ($encryptedFieldsMap as $ns => $encryptedFields) {
+                $encryptedFieldsMap[$ns] = json_decode(PackedArray::fromPHP($encryptedFields['fields'])->toRelaxedExtendedJSON(), true);
```
alcaeus, line 80: I presume this back and forth is to ensure that we've got a PHP representation of relaxed extended JSON that the dumper can then dump in the selected format?

src/Command/DumpEncryptedFieldsMapCommand.php
GromNaN: We already read the `encryptedFieldsMap` from the server, even if 2 document metadata use the same namespace, we would read the same configuration twice. No need to merge here.

I've left this comment in the code for supporting inheritance: https://github.com/doctrine/mongodb-odm/pull/2779/files#diff-3a40abcd893b19b63305f32c1e4ce3faf235a386cc387035a47d14b7b836d6eaR89

src/Command/DumpEncryptedFieldsMapCommand.php
GromNaN: Exactly, we need a JSON representation of the BSON document. `min` and `max` options can be `UTCDateTime` or BSON numbers, and `keyId` is always a `BSON\Binary`.

GromNaN: Tracked in [PHPORM-371](https://jira.mongodb.org/browse/PHPORM-371)

src/Command/DumpEncryptedFieldsMapCommand.php
alcaeus: Ok, that makes sense, thanks for confirming.

src/Command/DumpEncryptedFieldsMapCommand.php
alcaeus: I assume the last comment was referring to the previous comment thread about merging fields when inheritance is used. Thanks for confirming the encode/decode behaviour.

alcaeus (approved):

docs/encryption.rst
GromNaN: BSON binary is not supported because I use `json_encode` on the value in `DoctrineMongoDBExtension`.

There is no function that is able to read a mix of BSON and RelaxedExtendedJSON in an array.

PR #906, opened 2025-07-07
Add `--purge-with-delete` option to the `doctrine:mongodb:fixtures:load` command

GromNaN:
Requires https://github.com/doctrine/data-fixtures/pull/544

[PHPORM-364](https://jira.mongodb.org/browse/PHPORM-364)

### DIFF
--- a/src/Command/LoadDataFixturesDoctrineODMCommand.php
+++ b/src/Command/LoadDataFixturesDoctrineODMCommand.php
@@ -8,6 +8,7 @@ use Doctrine\Bundle\MongoDBBundle\Loader\SymfonyFixturesLoaderInterface;
 use Doctrine\Bundle\MongoDBBundle\ManagerRegistry;
 use Doctrine\Common\DataFixtures\Executor\MongoDBExecutor;
 use Doctrine\Common\DataFixtures\Purger\MongoDBPurger;
+use Doctrine\ODM\MongoDB\DocumentManager;
 use Psr\Log\AbstractLogger;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
@@ -15,7 +16,9 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\ConfirmationQuestion;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
+use function assert;
 use function implode;
+use function method_exists;
 use function sprintf;
 
 /**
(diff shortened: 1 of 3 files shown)


src/Command/LoadDataFixturesDoctrineODMCommand.php
```diff
@@ -50,16 +54,28 @@ protected function configure(): void
...
         $ui = new SymfonyStyle($input, $output);
 
-        if ($input->isInteractive() && ! $input->getOption('append')) {
+        if ($input->getOption('purge-with-delete')) {
+            if ($input->getOption('append')) {
+                $ui->error('The --purge-with-delete option cannot be used with the --append option.');
+
+                return self::INVALID;
+            }
+        } elseif ($input->isInteractive() && ! $input->getOption('append')) {
```
stof: should this really be a `elseif` ? Using `--purge-with-delete` without `--append` in an interactive session should still ask for confirmation for purging. To me, it should be a `if`

GromNaN: I think that if we have the `--purge-with-delete` option, we are explicitly asking to purge, so there is no need to confirm. But for me, we need an other option for `--purge-with-drop` that skip the confirmation. 

That said, it would be better to have a single option, `--purge`, with an optional value: `--purge`, `--purge=drop`, `--purge=delete`. It would skip the confirmation.

stof: you can already skip the confirmation with `--no-interaction` (or its `-n` shortcut).

GromNaN: Sure, I find the new `--purge` option more explicit.

GromNaN: I reverted to `--purge-with-delete` that only change the purger setting.

alcaeus (approved):

mickverm:
Any chance this could be released? I cannot figure out how to load my MongoDB test fixtures without `--purge-with-delete` when using `doctrine/data-fixtures:2.1.0`.

```bash
$ php bin/console --env=test doctrine:mongodb:fixtures:load --no-interaction
  > purging database
In CreateCollection.php line 207:
                                                         
  Collection already exists. NS: mongo_test.events  
                                                         
doctrine:mongodb:fixtures:load [--group GROUP] [--append] [--dm DM]
```

GromNaN:
@mickverm I created https://github.com/doctrine/data-fixtures/issues/549 to track this issue.

What you're reporting doesn't sound like a bundle compatibility issue, but a regression on `doctrine/data-fixtures`. Releasing a new version of the bundle will not fix the issue. I think the release will be created once the [auto-encryption feature](https://github.com/doctrine/DoctrineMongoDBBundle/pull/897) is merged.

mickverm:
@GromNaN Thank you for the quick response and the creation of the regression ticket, much appreciated 👍

PR #915, opened 2025-08-25
[Encryption] Fix query option min/max type for exported encrypted fields map

GromNaN:
When running the command `bin/console doctrine:mongodb:encryption:dump-fields-map` the min/max option of field query must match the field type.

But when we read the field map from the collection metadata, the `long` values are automatically converted to `int`.

In this PR, in manipulate the exported extends JSON to ensure the option value has the expected type.

Same issue as https://github.com/doctrine/mongodb-odm/pull/2805

### DIFF
--- a/src/Command/EncryptionDumpFieldsMapCommand.php
+++ b/src/Command/EncryptionDumpFieldsMapCommand.php
@@ -89,14 +89,6 @@ final class EncryptionDumpFieldsMapCommand extends Command
                         if (isset($field['queries']['max']['$numberInt'])) {
                             $field['queries']['max'] = ['$numberLong' => $field['queries']['max']['$numberInt']];
                         }
-                    } elseif ($field['bsonType'] === 'decimal') {
-                        if (isset($field['queries']['min']['$numberDouble'])) {
-                            $field['queries']['min'] = ['$numberDecimal' => $field['queries']['min']['$numberDouble']];
-                        }
-
-                        if (isset($field['queries']['max']['$numberDouble'])) {
-                            $field['queries']['max'] = ['$numberDecimal' => $field['queries']['max']['$numberDouble']];
-                        }
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


src/Command/EncryptionDumpFieldsMapCommand.php
```diff
@@ -76,9 +76,32 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+                            $field['queries']['max'] = ['$numberLong' => $field['queries']['max']['$numberInt']];
+                        }
+                    } elseif ($field['bsonType'] === 'decimal') {
+                        if (isset($field['queries']['min']['$numberDouble'])) {
+                            $field['queries']['min'] = ['$numberDecimal' => $field['queries']['min']['$numberDouble']];
+                        }
+
+                        if (isset($field['queries']['max']['$numberDouble'])) {
+                            $field['queries']['max'] = ['$numberDecimal' => $field['queries']['max']['$numberDouble']];
+                        }
```
GromNaN: I will revert this. Decimal is not supported for QE.
```
mongocryptd error: Type 'decimal' is not a supported equality indexed type
```

```suggestion
```

src/Command/EncryptionDumpFieldsMapCommand.php
```diff
@@ -76,9 +76,24 @@ protected function execute(InputInterface $input, OutputInterface $output): int
                 continue;
             }
 
+            // The min/max query options must have the same type as the field.
+            // But the PHP driver always convert to "int" or "float" when the value fit in the range
             foreach ($encryptedFieldsMap as $ns => $encryptedFields) {
+                $fields = json_decode(PackedArray::fromPHP($encryptedFields['fields'])->toCanonicalExtendedJSON(), true);
```
jmikola, line 82: See: https://github.com/doctrine/DoctrineMongoDBBundle/pull/897#discussion_r2306153348

PR #918, opened 2025-09-04
Fix UniqueValidator with new `identifierFieldNames` parameter

### LINKED ISSUE #874: Unique validator does not work when validation groups are used
### Bug Report

|    Q                                        |   A
|-------------------------------------------- | ------
| Version                                     | 5.1.1
| Previous Version if the bug is a regression | 5.1.1

#### Summary

Recently, I upgraded my Symfony version from 6.3 to 7.2. I've noticed that the Unique Validator stopped working when validation groups were present. 

Code without groups, the validator catches the error and prevents the duplicate in the database:
```PHP
<?php

namespace App\Document;

use Doctrine\ODM\MongoDB\Mapping\Annotations as MongoDB;
use Doctrine\Bundle\MongoDBBundle\Validator\Constraints\Unique as MongoDBUnique;

#[MongoDB\Document]
#[MongoDBUnique(fields: ['email'])]
class User
{
    #[MongoDB\Id]
    protected string $id;

    #[MongoDB\Field(type: 'string')]
    protected string $email;
}

```

Validation groups in use, the validator is not executed, causing a duplicate in the database:
```PHP
<?php

namespace App\Document;

use Doctrine\ODM\MongoDB\Mapping\Annotations as MongoDB;
use Doctrine\Bundle\MongoDBBundle\Validator\Constraints\Unique as MongoDBUnique;

#[MongoDB\Document]
#[MongoDBUnique(fields: ['email'], groups: ['registration'])]
class User
{
    #[MongoDB\Id]
    protected string $id;

    #[MongoDB\Field(type: 'string')]
    protected string $email;
}
```

#### Current behavior

The Unique validator is not executed when the validation groups are present.

#### Expected behavior

The Unique validator should work with and without validation groups.

#### How to reproduce

Install Symfony 7.2 with `doctrine/mongodb-odm-bundle`. 
Create a document with validation groups (example above).
Configure the form to use the validation group:
```PHP

namespace App\Form\Type;

use App\Document\User;
use Symfony\Component\Form\AbstractType;
use Symfony\Component\Form\Extension\Core\Type\EmailType;
use Symfony\Component\Form\FormBuilderInterface;
use Symfony\Component\OptionsResolver\OptionsResolver;

class UserType extends AbstractType
{
    public function buildForm(FormBuilderInterface $builder, array $options): void
    {
        $builder
            ->add('email', EmailType::class)
        ;
    }

    public function configureOptions(OptionsResolver $resolver): void
    {
        $resolver->setDefaults([
            'data_class' => User::class,
            'validation_groups' => ['registration'],
        ]);
    }
}

```
Handle form submission in the controller:
```PHP
$user = new User();
$form = $this->createForm(UserType::class, $user);
$form->handleRequest($request);
if ($form->isSubmitted() && $form->isValid()) {
     $dm->persist($user);
     $dm->flush();
}
```
Try to submit the form twice using the same email. It will create two documents with the same email, but it should display a validation error on the second attempt.

soyuka:
Fix https://github.com/doctrine/DoctrineMongoDBBundle/issues/874

Breaking change in Symfony https://github.com/symfony/symfony/pull/38662


GromNaN (approved):

GromNaN:
I completely revised the implementation. In fact, we don't have to override the constructor to change only the default value of the `$service` parameter. We only need to redeclare the default value in the `$service` property.

We have to declare the class conditionally to keep the compatibility with Symfony 6.4 as the prop type has been [added in 7.0](https://github.com/symfony/symfony/commit/7ea24615d7b6f46a2af42cf860cd5f74f38eb7ad#diff-580bed00c9b884da104fb61d62ee451c36ef1ec4936c5a55b68aa4bd182c701b).

src/Validator/Constraints/Unique.php
```diff
@@ -5,46 +5,36 @@
...
-/**
- * Constraint for the unique document validator
- *
- * @Annotation
- * @Target({"CLASS", "ANNOTATION"})
- */
-#[Attribute(Attribute::TARGET_CLASS | Attribute::IS_REPEATABLE)]
-class Unique extends UniqueEntity
-{
+if ((new ReflectionProperty(UniqueEntity::class, 'service'))->hasType()) {
```
jmikola, line 11: Noted that this is essentially a test for Symfony 7.0+ based on [symfony/symfony@7ea2461](https://github.com/symfony/symfony/commit/7ea24615d7b6f46a2af42cf860cd5f74f38eb7ad#diff-580bed00c9b884da104fb61d62ee451c36ef1ec4936c5a55b68aa4bd182c701b) adding parameter types to UniqueEntity.

tests/Validator/Constraints/UniqueTest.php
```diff
@@ -23,11 +25,71 @@ public function testWithDefaultProperty(): void
         self::assertInstanceOf(Unique::class, $constraint);
         self::assertSame(['email'], $constraint->fields);
         self::assertSame('doctrine_odm.mongodb.unique', $constraint->validatedBy());
+        self::assertNull($constraint->em);
```
jmikola, line 28: I'm not questioning the correctness here, but why was a `$constraint->em` assertion introduced?

GromNaN: I hit an issue while working on the previous implementation.

GromNaN: It would have been so nice if this parameter was named `manager` to take into account the ODM.

jmikola (approved):

GromNaN:
Thank you @soyuka for raising the issue again and proposing a solution.

soyuka:
thanks @GromNaN for taking this over my dirty patch :D

PR #921, opened 2025-09-21
Use `ArrayNodeDefinition::acceptAndWrap`

GromNaN:
Enable more specific types.

- [ ] Requires https://github.com/symfony/symfony/pull/51273
- [ ] Use the new method conditionaly to keep support for Symfony < 7.4

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -53,17 +53,8 @@ class Configuration implements ConfigurationInterface
                 ->scalarNode('auto_generate_proxy_classes')
                     ->defaultValue(ODMConfiguration::AUTOGENERATE_EVAL)
                     ->beforeNormalization()
-                    ->always(static function ($v) {
-                        if ($v === false) {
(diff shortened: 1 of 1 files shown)


src/DependencyInjection/Configuration.php
```diff
@@ -51,42 +51,26 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->end()
                 ->end()
                 ->scalarNode('auto_generate_proxy_classes')
+                //->integerNode('auto_generate_proxy_classes')
```
GromNaN: By using `ifTrue` and `ifFalse`, boolean should be allowed.
Error with the `intergetNode`:

```
Invalid type for path "doctrine_mongodb.auto_generate_proxy_classes". Expected "int", but got "bool".
src/Symfony/Component/Config/Definition/IntegerNode.php:26
```

nicolas-grekas: I'm not sure using integer node was a good idea: it plays against using env vars...

PR #923, opened 2025-10-08
Inline `AbstractDoctrineExtension` from Symfony Doctrine bridge

greg0ire:
In order to be make it easier to make changes in the metadata discovery mechanism, let us avoid having to change code in 2 different repositories and inline the code of the AbstractDoctrineExtension of the Symfony bridge here. This class will be deprecated and ultimately, removed.

No type hints have been added for the sake of backward compatibility, but there is still a breaking change: the extension no longer pass an instanceof AbstractDoctrineExtension assertion, and I think that's fine.

### DIFF
--- a/src/DependencyInjection/DoctrineMongoDBExtension.php
+++ b/src/DependencyInjection/DoctrineMongoDBExtension.php
@@ -27,10 +27,11 @@ use Doctrine\ODM\MongoDB\Mapping\Driver\AttributeDriver;
 use Doctrine\Persistence\Mapping\Driver\MappingDriverChain;
 use Doctrine\Persistence\Proxy;
 use InvalidArgumentException;
+use LogicException;
 use MongoDB\BSON\Document as BsonDocument;
 use MongoDB\Client;
 use ProxyManager\Proxy\LazyLoadingInterface;
-use Symfony\Bridge\Doctrine\DependencyInjection\AbstractDoctrineExtension;
+use ReflectionClass;
 use Symfony\Bridge\Doctrine\Messenger\DoctrineClearEntityManagerWorkerSubscriber;
 use Symfony\Component\Cache\Adapter\ApcuAdapter;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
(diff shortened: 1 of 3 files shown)


src/DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -679,12 +1036,85 @@ public function getXsdValidationBasePath(): string
...
+                $namespace = 'sf_' . $this->getMappingResourceExtension() . '_' . $objectManagerName . '_' . ContainerBuilder::hash($seed);
+
+                $cacheDriver['namespace'] = $namespace;
+            }
+
+            $cacheDef->addMethodCall('setNamespace', [$cacheDriver['namespace']]);
+
+            $container->setDefinition($cacheDriverServiceId, $cacheDef);
+
+            return $cacheDriverServiceId;
```
greg0ire, line 1133: This part of the code is the one responsible for the drop in code coverage, but should get removed in the future.

stof (approved):

src/DependencyInjection/DoctrineMongoDBExtension.php
stof: Remaining code related to support of `doctrine/cache` v1 should be deprecated entirely (which will remove this code once removed)

src/DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -27,49 +27,466 @@
...
+    ) {
+        if (! $mappingConfig['type'] || ! $mappingConfig['dir'] || ! $mappingConfig['prefix']) {
+            throw new InvalidArgumentException(sprintf('Mapping definitions for Doctrine manager "%s" require at least the "type", "dir" and "prefix" options.', $objectManagerName));
+        }
+
+        if (! is_dir($mappingConfig['dir'])) {
+            throw new InvalidArgumentException(sprintf('Specified non-existing directory "%s" as Doctrine mapping source.', $mappingConfig['dir']));
+        }
+
+        if (! in_array($mappingConfig['type'], ['xml', 'yml', 'annotation', 'php', 'staticphp', 'attribute'])) {
```
GromNaN: annotation support was removed from the MongoDB Bundle in [v5.0.0](https://github.com/doctrine/DoctrineMongoDBBundle/releases/tag/5.0.0) by https://github.com/doctrine/DoctrineMongoDBBundle/pull/793

It is not necessary to reimport all the logic to support them.

greg0ire: OK, I removed it.

src/DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -45,31 +46,404 @@
...
+            } else {
+                $mappingDriverDef = new Definition($this->getMetadataDriverClass($driverType), [
+                    array_values($driverPaths),
+                ]);
+            }
+
+            if (
+                str_contains($mappingDriverDef->getClass(), 'yml') || str_contains($mappingDriverDef->getClass(), 'xml')
+                || str_contains($mappingDriverDef->getClass(), 'Yaml') || str_contains($mappingDriverDef->getClass(), 'Xml')
+            ) {
```
GromNaN: Yaml is not supported since MongoDB ODM v2.
This bundle removed usage of `.class` parameters, we need to check only for `Xml` in the class name.

```suggestion
            if (str_contains($mappingDriverDef->getClass(), 'Xml')) {
```

src/DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -45,31 +46,404 @@
...
+    ) {
+        if (! $mappingConfig['type'] || ! $mappingConfig['dir'] || ! $mappingConfig['prefix']) {
+            throw new InvalidArgumentException(sprintf('Mapping definitions for Doctrine manager "%s" require at least the "type", "dir" and "prefix" options.', $objectManagerName));
+        }
+
+        if (! is_dir($mappingConfig['dir'])) {
+            throw new InvalidArgumentException(sprintf('Specified non-existing directory "%s" as Doctrine mapping source.', $mappingConfig['dir']));
+        }
+
+        if (! in_array($mappingConfig['type'], ['xml', 'yml', 'php', 'staticphp', 'attribute'])) {
```
GromNaN: Yaml not supported.
```suggestion
        if (! in_array($mappingConfig['type'], ['xml', 'php', 'staticphp', 'attribute'])) {
```

src/DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -45,31 +46,404 @@
...
+    {
+        $configPath = $this->getMappingResourceConfigDirectory($dir);
+        $extension  = $this->getMappingResourceExtension();
+
+        if (glob($dir . '/' . $configPath . '/*.' . $extension . '.xml', GLOB_NOSORT)) {
+            $driver = 'xml';
+        } elseif (glob($dir . '/' . $configPath . '/*.' . $extension . '.yml', GLOB_NOSORT)) {
+            $driver = 'yml';
+        } elseif (glob($dir . '/' . $configPath . '/*.' . $extension . '.php', GLOB_NOSORT)) {
+            $driver = 'php';
```
GromNaN, line 344: Yaml format not supported by MongoDB ODM. I'm not sure for the PHP format.
```suggestion
        if (glob($dir . '/' . $configPath . '/*.' . $extension . '.xml', GLOB_NOSORT)) {
            $driver = 'xml';
        } elseif (glob($dir . '/' . $configPath . '/*.' . $extension . '.php', GLOB_NOSORT)) {
            $driver = 'php';
```

GromNaN (approved):
Code related to Yaml can be removed, we can do it in another PR to keep things clear.

GromNaN (approved):
Thank you @greg0ire.

PR #931, opened 2025-10-17
Make `MONGODB_URI` configurable for functional tests

GromNaN:


tests/Command/CommandTestKernel.php
```diff
@@ -44,7 +44,7 @@ protected function configureContainer(ContainerBuilder $container, LoaderInterfa
         ]);
 
         $container->loadFromExtension('doctrine_mongodb', [
-            'connections' => ['default' => []],
+            'connections' => ['default' => ['server' => '%env(DOCTRINE_MONGODB_SERVER)%']],
```
jmikola: Should this have a fallback value as is done in the base [TestCase](https://github.com/doctrine/DoctrineMongoDBBundle/blob/5.4.2/tests/TestCase.php#L30) class? I'm not familiar with how `%env()%` resolution works exactly (I did consult the [Symfony docs](https://symfony.com/doc/current/configuration.html#configuration-based-on-environment-variables)), but I imagine this is going to resolve to an empty string if `DOCTRINE_MONGODB_SERVER` is undefined. That seems less desirable than `null`, which would allow the library to apply its default value.

An empty string might yield an InvalidArgumentException with:

> Failed to parse MongoDB URI: ''. Invalid URI, no scheme part specified.

GromNaN: This is only for tests, and the value is set in PHPUnit configuration. I don't find the need for a fallback here. https://github.com/doctrine/DoctrineMongoDBBundle/blob/20a6556f5b3cb7a1dacca4dcfd5750bb8a6dffee/phpunit.xml.dist#L23

PR #940, opened 2025-10-29
Enable native lazy objects by default, add an option to disable it

GromNaN:
Use native lazy objects by default when `PHP >= 7.4` and `doctrine/mongodb-odm >= 2.14` (https://github.com/doctrine/mongodb-odm/pull/2840)
Add a new option `enable_native_lazy_objects` to disable this behavior.

Todo
- [x] Deprecate not using native lazy object when supported
- [x] Make `AbstractManagerRegistry::$proxyInterfaceName` nullable https://github.com/doctrine/persistence/pull/458

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -635,7 +635,7 @@ Lazy object implementation
 Doctrine MongoDB ODM uses lazy objects for lazy instantiation of references.
 The original implementation is based on the `ProxyManager` library.
 Since version 2.10 of Doctrine MongoDB ODM, support for Symfony lazy ghost
-objects has been added. And in version 2.14, support for PHP 7.4 native lazy
+objects has been added. And in version 2.14, support for PHP 8.4 native lazy
 objects has been added.
 
 The bundle select the best available lazy object implementation based on the
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -629,6 +629,47 @@ Using Queryable Encryption
 
 For details on configuring Queryable Encryption (QE) and Client-Side Field-Level Encryption (CSFLE), see :doc:`encryption`.
 
+Lazy object implementation
+--------------------------
+
+Doctrine MongoDB ODM uses lazy objects for lazy instantiation of references.
+The original implementation is based on the `ProxyManager` library.
+Since version 2.10 of Doctrine MongoDB ODM, support for Symfony lazy ghost
+objects has been added. And in version 2.14, support for PHP 7.4 native lazy
+objects has been added.
+
+The bundle select the best available lazy object implementation based on the
+installed packages and PHP version. You can override this behavior by setting
+the following configuration options to enable or disable specific lazy object
+implementations. This is not recommended unless you have a specific reason to do
+so. Please open an issue if the default Native Lazy Objects are not working as
+expected.
+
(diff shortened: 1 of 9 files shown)


docs/config.rst
```diff
@@ -629,6 +629,47 @@ Using Queryable Encryption
 
 For details on configuring Queryable Encryption (QE) and Client-Side Field-Level Encryption (CSFLE), see :doc:`encryption`.
 
+Lazy object implementation
+--------------------------
+
+Doctrine MongoDB ODM uses lazy objects for lazy instantiation of references.
+The original implementation is based on the `ProxyManager` library.
+Since version 2.10 of Doctrine MongoDB ODM, support for Symfony lazy ghost
+objects has been added. And in version 2.14, support for PHP 7.4 native lazy
```
mickverm: Shouldn't this be 8.4? You mentioned 7.4 in the PR description as well.

src/DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -486,9 +486,8 @@ public function load(array $configs, ContainerBuilder $container): void
             $container->removeDefinition('doctrine_mongodb.odm.command.load_data_fixtures');
         }
 
-        // Requires doctrine/mongodb-odm 2.10
         $container->getDefinition('doctrine_mongodb')
-            ->setArgument(5, $config['enable_lazy_ghost_objects'] ? Proxy::class : LazyLoadingInterface::class);
+            ->setArgument(5, $config['enable_lazy_ghost_objects'] ? LazyLoadingInterface::class : Proxy::class);
```
alcaeus, line 490: Is this a bug in the previous implementation, or why does this need changing?

docs/config.rst
GromNaN: Good catch. Thanks.

src/DependencyInjection/DoctrineMongoDBExtension.php
GromNaN: It looks like a bug in the previous implementation. Not covered by a test.

docs/config.rst
```diff
@@ -629,6 +629,46 @@ Using Queryable Encryption
...
+The original implementation is based on the `ProxyManager` library.
+In version 2.10, support for Symfony lazy ghost objects has been added.
+And in version 2.14, support for PHP 8.4 native lazy objects has been added.
+
+The bundle selects the best available lazy object implementation based on the
+installed packages and PHP version. You can override this behavior by setting
+the following configuration options disable specific lazy object implementations.
+This is not recommended unless you have a specific reason to do so.
+Please open an issue if the default Native Lazy Objects are not working as expected
+as it is the preferred implementation, other will be removed in future versions.
```
alcaeus:
```suggestion
as it is the preferred implementation. The other implementations will be removed in a future version.
```

alcaeus (approved):
Minor wording change, but LGTM!

src/DependencyInjection/DoctrineMongoDBExtension.php
GromNaN: I was wrong. The `LazyLoadingInterface` is for ProxyManager.

PR #938, opened 2025-10-28
Enable support for Symfony 8

GromNaN:
Rework the test matrix to test all PHP versions with latest composer versions, and all Symfony versions with PHP 8.4

PR #941, opened 2025-10-29
Run tests with Symfony Messenger

GromNaN:
Currently, integration with `symfony/messenger` is not tested because the dependency is not loaded.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,6 +47,7 @@
         "squizlabs/php_codesniffer": "^4",
         "symfony/browser-kit": "^6.4 || ^7.0",
         "symfony/form": "^6.4 || ^7.0",
+        "symfony/messenger": "^6.4 || ^7.0",
         "symfony/phpunit-bridge": "^6.4.1 || ^7.0.1",
         "symfony/security-bundle": "^6.4 || ^7.0",
         "symfony/stopwatch": "^6.4 || ^7.0",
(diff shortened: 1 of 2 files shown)


tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
```diff
@@ -363,14 +359,6 @@ public function testPublicServicesAndAliases(): void
 
     public function testMessengerIntegration(): void
     {
-        if (! interface_exists(MessageBusInterface::class)) {
-            $this->markTestSkipped('Symfony Messenger component is not installed');
-        }
-
-        if (! class_exists(DoctrineClearEntityManagerWorkerSubscriber::class)) {
-            $this->markTestSkipped('DoctrineClearEntityManagerWorkerSubscriber is not available in symfony/doctrine-bridge');
```
GromNaN, line 371: Available since Symfony 4.4

composer.json
```diff
@@ -47,6 +47,7 @@
         "squizlabs/php_codesniffer": "^4",
         "symfony/browser-kit": "^6.4 || ^7.0",
         "symfony/form": "^6.4 || ^7.0",
+        "symfony/messenger": "^6.4 || ^7.0",
```
alcaeus, line 50: Short reminder that this will conflict with #938 and need changing.

alcaeus (approved):

PR #943, opened 2025-11-05
kernel.reset tag to doctrine_mongodb definition

soyuka:
Fixes https://github.com/php/frankenphp/issues/1767

the `ResetInterface` is not autoconfigured for bundles therefore this is never triggered by the ServicesResetter.

### DIFF
--- a/config/mongodb.php
+++ b/config/mongodb.php
@@ -56,6 +56,7 @@ return static function (ContainerConfigurator $containerConfigurator): void {
                 abstract_arg('Proxy Interface Name'),
                 service('service_container'),
             ])
+            ->tag('kernel.reset', ['method' => 'reset'])
 
         ->set('doctrine_mongodb.odm.listeners.resolve_target_document', ResolveTargetDocumentListener::class)
 


src/DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -115,7 +115,8 @@ public function load(array $configs, ContainerBuilder $container): void
 
         // Requires doctrine/mongodb-odm 2.10
         $container->getDefinition('doctrine_mongodb')
-            ->setArgument(5, $config['enable_lazy_ghost_objects'] ? Proxy::class : LazyLoadingInterface::class);
+            ->setArgument(5, $config['enable_lazy_ghost_objects'] ? Proxy::class : LazyLoadingInterface::class)
+            ->addTag('kernel.reset', ['method' => 'reset']);
```
GromNaN: It did not work if you add the tag in the config file?
https://github.com/doctrine/DoctrineMongoDBBundle/blob/062b66e744dd3cd0a12609469c9e9307ed7fa16f/config/mongodb.php#L48

GromNaN (approved):

GromNaN:
Thank you @soyuka

PR #947, opened 2025-11-14
Fix proxy interface for ProxyManager or Symfony LazyGhost

GromNaN:
I was wrong in https://github.com/doctrine/DoctrineMongoDBBundle/pull/940#discussion_r2527440188.

Fix the interface that identifies the lazy objects.

Note that `Doctrine\Persistence\Mapping\ProxyClassNameResolver` should be used by `Doctrine\Persistence\AbstractManagerRegistry` instead of repeating the same logic.

### DIFF
--- a/src/DependencyInjection/DoctrineMongoDBExtension.php
+++ b/src/DependencyInjection/DoctrineMongoDBExtension.php
@@ -487,7 +487,7 @@ class DoctrineMongoDBExtension extends Extension
         }
 
         $container->getDefinition('doctrine_mongodb')
-            ->setArgument(5, $config['enable_lazy_ghost_objects'] ? LazyLoadingInterface::class : Proxy::class);
+            ->setArgument(5, $config['enable_lazy_ghost_objects'] ? Proxy::class : LazyLoadingInterface::class);
 
         // load the connections
         $this->loadConnections($config['connections'], $container, $config);
--- a/tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
+++ b/tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
@@ -777,13 +777,13 @@ class DoctrineMongoDBExtensionTest extends TestCase
             ];
         }
 
-        if (trait_exists(LazyGhostTrait::class)) {
+        if (trait_exists(LazyGhostTrait::class) && method_exists(Configuration::class, 'setUseLazyGhostObject')) {
             yield 'Symfony Lazy Objects' => [
                 ['enable_lazy_ghost_objects' => true, 'enable_native_lazy_objects' => false],
             ];
         }
 
-        if (PHP_VERSION_ID >= 80400) {
+        if (PHP_VERSION_ID >= 80400 && method_exists(Configuration::class, 'setUseNativeLazyObject')) {
             yield 'Native Lazy Objects' => [
                 ['enable_lazy_ghost_objects' => false, 'enable_native_lazy_objects' => true],
             ];
--- a/tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
+++ b/tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
@@ -758,7 +758,7 @@ class DoctrineMongoDBExtensionTest extends TestCase
         // Create a lazy object
         $ref = $dm->getReference(TestDocument::class, 'some');
         self::assertInstanceOf(TestDocument::class, $ref);
-        self::assertSame($dm, $registry->getManagerForClass($ref::class));
+        self::assertSame($dm, $registry->getManagerForClass($ref::class), 'The manager is found for the proxy document class');
     }
 
     public static function provideLazyObjectConfigurations(): iterable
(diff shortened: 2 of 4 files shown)


src/DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -487,7 +487,7 @@ public function load(array $configs, ContainerBuilder $container): void
         }
 
         $container->getDefinition('doctrine_mongodb')
-            ->setArgument(5, $config['enable_lazy_ghost_objects'] ? LazyLoadingInterface::class : Proxy::class);
+            ->setArgument(5, $config['enable_lazy_ghost_objects'] ? Proxy::class : LazyLoadingInterface::class);
```
jmikola, line 490: Reading https://github.com/doctrine/DoctrineMongoDBBundle/pull/940#discussion_r2527440188 I'm still confused. These aren't the best class names. Is it worth adding a block comment above this to explain?

I went so far as to look at https://github.com/FriendsOfPHP/proxy-manager-lts/blob/1.x/src/ProxyManager/Proxy/LazyLoadingInterface.php and it's not even clear from that file.

tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
```diff
@@ -716,6 +727,61 @@ public function testAutoEncryptionMinimumODMVersion(): void
         $loader->load([$config], $container);
     }
 
+    #[DataProvider('provideLazyObjectConfigurations')]
```
jmikola, line 737: Does PHPStan want you to annotate `array $config` as an iterable of array shapes? I don't think that's necessary.

tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
```diff
@@ -716,6 +727,61 @@ public function testAutoEncryptionMinimumODMVersion(): void
...
+
+        if (trait_exists(LazyGhostTrait::class)) {
+            yield 'Symfony Lazy Objects' => [
+                ['enable_lazy_ghost_objects' => true, 'enable_native_lazy_objects' => false],
+            ];
+        }
+
+        if (PHP_VERSION_ID >= 80400) {
+            yield 'Native Lazy Objects' => [
+                ['enable_lazy_ghost_objects' => false, 'enable_native_lazy_objects' => true],
```
jmikola, line 788: Should you test specifying `true` for both options? I assume `enable_native_lazy_objects` might take precedence, but would understand if they are mutually exclusive and should throw.

tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
```diff
@@ -716,6 +727,61 @@ public function testAutoEncryptionMinimumODMVersion(): void
...
+        $registry = $container->get('doctrine_mongodb');
+        $dm       = $container->get('doctrine_mongodb.odm.document_manager');
+
+        self::assertInstanceOf(ManagerRegistry::class, $registry);
+        self::assertInstanceOf(DocumentManager::class, $dm);
+
+        // Create a lazy object
+        $ref = $dm->getReference(TestDocument::class, 'some');
+        self::assertInstanceOf(TestDocument::class, $ref);
+        self::assertSame($dm, $registry->getManagerForClass($ref::class));
```
GromNaN:
```suggestion
        self::assertSame($dm, $registry->getManagerForClass($ref::class), 'The manager is found for the proxy document class');
```

tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
```diff
@@ -716,6 +727,61 @@ public function testAutoEncryptionMinimumODMVersion(): void
...
+        (new CreateProxyDirectoryPass())->process($container);
+
+        $container->compile();
+        $registry = $container->get('doctrine_mongodb');
+        $dm       = $container->get('doctrine_mongodb.odm.document_manager');
+
+        self::assertInstanceOf(ManagerRegistry::class, $registry);
+        self::assertInstanceOf(DocumentManager::class, $dm);
+
+        // Create a lazy object
```
jmikola, line 765: Consider a bigger comment that explains the last assertion on the DM registry is actually pertinent to testing the 5th argument (and your bug fix).

Assertion messages also work fine.

jmikola (approved):

tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
GromNaN: `enable_native_lazy_objects` has priority over `enable_lazy_ghost_objects`. I don't test which lazy object implementation is used here, only overriding both default values since they are dynamically set depending of the PHP and ODM versions.

tests/DependencyInjection/DoctrineMongoDBExtensionTest.php
GromNaN: I added the `@phpstan-type`

src/DependencyInjection/DoctrineMongoDBExtension.php
GromNaN: This interfaces are used to get the parent class when looking for a class metadata from a proxy class name.
https://github.com/doctrine/persistence/blob/00bd9ccc1f9c6a74fc2dc557313869acd1998cba/src/AbstractManagerRegistry.php#L135

PR #948, opened 2025-11-14
Clean supported metadata driver types

GromNaN:
After https://github.com/doctrine/DoctrineMongoDBBundle/pull/923, we can cleanup some legacy.
- The `yml` format isn't supported since MongoDB ODM v2
- The `php` and `staticphp` format have never been supported by the bundle itself. They require a custom driver configuration (see https://github.com/doctrine/DoctrineMongoDBBundle/pull/936)

### DIFF
--- a/src/DependencyInjection/DoctrineMongoDBExtension.php
+++ b/src/DependencyInjection/DoctrineMongoDBExtension.php
@@ -235,7 +235,7 @@ class DoctrineMongoDBExtension extends Extension
         }
 
         if (! $bundleConfig['dir']) {
-            if (in_array($bundleConfig['type'], ['staticphp', 'attribute'])) {
+            if (in_array($bundleConfig['type'], ['attribute'])) {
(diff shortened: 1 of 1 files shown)


src/DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -325,8 +321,8 @@ protected function assertValidMappingConfiguration(
             throw new InvalidArgumentException(sprintf('Specified non-existing directory "%s" as Doctrine mapping source.', $mappingConfig['dir']));
         }
 
-        if (! in_array($mappingConfig['type'], ['xml', 'php', 'staticphp', 'attribute'])) {
```
GromNaN, line 328: Using `php` or `static` php currently fails in https://github.com/doctrine/DoctrineMongoDBBundle/blob/a4279b14c9a1ec66de86f29259569144fbfbd76f/src/DependencyInjection/DoctrineMongoDBExtension.php#L1021-L1029

jmikola (approved):

PR #950, opened 2025-11-14
Update .doctrine-project.json for 5.5.0 and remove old branches

GromNaN:
Currently only branches from 5.x have the documentation published.

image

But the version 4.7 is still very used:
image

### DIFF
--- a/.doctrine-project.json
+++ b/.doctrine-project.json
@@ -4,105 +4,30 @@
     "slug": "doctrine-mongodb-bundle",
     "versions": [
         {
-            "name": "5.5",
-            "branchName": "5.5.x",
-            "slug": "5.5",
+            "name": "5.6",
+            "branchName": "5.6.x",
+            "slug": "latest",
             "upcoming": true
         },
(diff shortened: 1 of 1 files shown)


.doctrine-project.json
```diff
@@ -3,107 +3,28 @@
...
-            "branchName": "5.0.x",
-            "slug": "5.0",
-            "maintained": false
-        },
-        {
-            "name": "4.7",
-            "branchName": "4.7.x",
-            "slug": "4.7",
-            "maintained": false
-        },
```
greg0ire, line 47: Since 4.7 is still very used, maybe we should keep this block, but remove the `branchName`, so that the docs are generated from the latest tag?

GromNaN: > so that the docs are generated from the latest tag?

I didn't know that.
Version reverted.

greg0ire (approved):

greg0ire:
Thanks @GromNaN !

PR #951, opened 2025-11-19
Replace deprecated `Symfony\Component\HttpKernel\DependencyInjection\Extension`

mickverm:
I noticed the following deprecation warning in our project:

```
User Deprecated: The "Symfony\Component\HttpKernel\DependencyInjection\Extension" class is considered internal since Symfony 7.1, to be deprecated in 8.1; use Symfony\Component\DependencyInjection\Extension\Extension instead. It may change without further notice. You should not use it from "Doctrine\Bundle\MongoDBBundle\DependencyInjection\DoctrineMongoDBExtension".
```


GromNaN (approved):
Good catch. Thank you!

mickverm:
That was fast 🥳

PR #953, opened 2025-11-20
Test with PHP 8.5

GromNaN:


IonBazan (approved):
Deprecations triggered:
```
Method ReflectionProperty::setAccessible() is deprecated since 8.5, as it has no effect
- tests/CacheWarmer/ProxyCacheWarmerTest.php:39
- tests/Mapping/Driver/AbstractDriverTestCase.php:15

Using null as an array offset is deprecated, use an empty string instead
- vendor/doctrine/mongodb-odm/src/Persisters/DocumentPersister.php:997
```

See: https://github.com/doctrine/mongodb-odm/blob/1a86e7d6e8f3cea89e9add0d78cf6915ddbc6793/src/Persisters/DocumentPersister.php#L997

PR #946, opened 2025-11-14
Disable proxy class generation for PHP 8.4 native lazy objects

GromNaN:

PR #963, opened 2025-12-10
Add  `#[AsFieldType]` to custom field types as services

GromNaN:
Bundle implementation of the `TypeRegistry` from https://github.com/doctrine/mongodb-odm/pull/2966

In order to preserve BC in case people manipulate types with the methods such as `Type::add()`, we use the singleton `TypeRegistry::getSharedInstance()`.

The new option `doctrine_mongodb.type_registry` enables a single instance of `TypeRegistry` per `DocumentManager`, and the autoconfiguration of services with the `#[AsFieldType]` attribute.

```php
#[AsFieldType(Money::class, objectManager: 'customer_manager')]
final class MoneyType extends Type
{
    // ...
}
```

This feature is very powerful, combined with the detection of the Field type from the property type.

```php
use Money\Money;
use Doctrine\ODM\MongoDB\Mapping\Annotations as ODM;

#[ODM\Document]
class Product
{
    #[ODM\Id]
    public ?string $id = null;

    #[ODM\Field]
    public Money $price;
}
```

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -317,9 +317,11 @@ your documents.
             </doctrine_mongodb:config>
         </container>
 
-You can also use a service as custom type by providing the service id. It's
-necessary to disable the shared type registry by setting ``share_type_registry``
-to ``false`` in this case.
+You can register custom types by adding the tag ``doctrine_mongodb.odm.field_type``
+to services with the following parameters:
+ - ``type`` is the name of the type. Must be unique.
+ - ``object_manager`` (optional) name of the document manager where the type
+   will be registered. If not set, the type will be registered in all document managers.
 
 .. configuration-block::
 
@@ -330,7 +332,7 @@ to ``false`` in this case.
                 class: Fully\Qualified\Class\Name
 
         doctrine_mongodb:
-            share_type_registry: false
+            type_registry: true
             types:
                 custom_type: @app.custom_type_service
 
@@ -349,7 +351,7 @@ to ``false`` in this case.
             </services>
 
             <doctrine_mongodb:config
-                share-type-registry="false"
+                type-registry="true"
             >
                 <doctrine_mongodb:type name="custom_type" service="app.custom_type_service" />
             </doctrine_mongodb:config>
@@ -358,6 +360,10 @@ to ``false`` in this case.
 .. note::
 
     Defining a custom type as a service requires Doctrine MongoDB ODM 2.16 or higher.
+
+You can also register services as custom types by tagging with the
+``#[AsFieldType]`` attribute and enabling autoconfiguration.
+See `Registering Custom Field Types` for more details.
 
 Filters
 -------
@@ -980,6 +986,7 @@ Full Default Configuration
         };
 
 .. _`Custom types`: https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/current/reference/custom-mapping-types.html
+.. _`Registering Custom Field Types`: cookbook/field_type
 .. _`define it as an environment variable`: https://symfony.com/doc/current/configuration.html#configuration-based-on-environment-variables
 .. _`connection string`: https://docs.mongodb.com/manual/reference/connection-string/#urioption.authSource
 .. _`Replica Sets`: https://www.php.net/manual/en/mongo.connecting.rs.php
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -317,15 +317,47 @@ your documents.
             </doctrine_mongodb:config>
         </container>
 
-    .. code-block:: php
+You can also use a service as custom type by providing the service id. It's
+necessary to disable the shared type registry by setting ``share_type_registry``
+to ``false`` in this case.
 
-        use Symfony\Config\DoctrineMongodbConfig;
+.. configuration-block::
 
-        return static function (DoctrineMongodbConfig $config): void {
-            $config->type('custom_type')
-                ->class(\Fully\Qualified\Class\Name::class)
-            ;
-        }
+    .. code-block:: yaml
+
+        services:
+            app.custom_type_service:
+                class: Fully\Qualified\Class\Name
+
+        doctrine_mongodb:
+            share_type_registry: false
+            types:
+                custom_type: @app.custom_type_service
+
+    .. code-block:: xml
+
+        <?xml version="1.0" ?>
+
+        <container xmlns="http://symfony.com/schema/dic/services"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xmlns:doctrine_mongodb="http://symfony.com/schema/dic/doctrine/odm/mongodb"
+                   xsi:schemaLocation="http://symfony.com/schema/dic/services https://symfony.com/schema/dic/services/services-1.0.xsd
+                                        http://symfony.com/schema/dic/doctrine/odm/mongodb https://symfony.com/schema/dic/doctrine/odm/mongodb/mongodb-1.0.xsd">
+
+            <services>
+                <service id="app.custom_type_service" class="Fully\Qualified\Class\Name" />
+            </services>
+
+            <doctrine_mongodb:config
+                share-type-registry="false"
+            >
+                <doctrine_mongodb:type name="custom_type" service="app.custom_type_service" />
+            </doctrine_mongodb:config>
+        </container>
+
+.. note::
+
+    Defining a custom type as a service requires Doctrine MongoDB ODM 2.16 or higher.
 
 Filters
 -------
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -332,7 +332,7 @@ to services with the following parameters:
                 class: Fully\Qualified\Class\Name
 
         doctrine_mongodb:
-            type_registry: true
+            scoped_type_registry: true
             types:
                 custom_type: @app.custom_type_service
 
--- a/docs/cookbook/field_type.rst
+++ b/docs/cookbook/field_type.rst
@@ -8,13 +8,16 @@ specific data formats or structures.
 Enable the Type Registry
 ------------------------
 
-The ``TypeRegistry`` is responsible for managing custom field types in Doctrine
-ODM. To use it, you need to enable it in the configuration:
+The ``TypeRegistry`` is responsible for managing field types in Doctrine MongoDB
+ODM. By default, there is a global type registry shared across all document
+managers. However, you can enable a scoped type registry for each document manager
+and using service autoconfiguration by setting the ``scoped_type_registry``
+option to ``true``:
 
 .. code-block:: yaml
 
     doctrine_mongodb:
-        type_registry: true
+        scoped_type_registry: true
 
 Creating a Custom Field Type
 ----------------------------
--- a/docs/cookbook/field_type.rst
+++ b/docs/cookbook/field_type.rst
@@ -1,0 +1,131 @@
+Registering Custom Field Types
+==============================
+
+Sometimes, the built-in field types provided by Doctrine ODM are not sufficient
+for your application. In such cases, you can create `Custom Mapping Types`_ to handle
+specific data formats or structures.
+
+Enable the Type Registry
+------------------------
+
+The ``TypeRegistry`` is responsible for managing custom field types in Doctrine
+ODM. To use it, you need to enable it in the configuration:
+
+.. code-block:: yaml
+
+    doctrine_mongodb:
+        type_registry: true
+
+Creating a Custom Field Type
+----------------------------
+
+In order to create a custom field type, you need to extend the
+``Doctrine\ODM\MongoDB\Types\Type`` class and implement the required methods.
+
+Mapping a Money Value Object
+----------------------------
+
+You can create a custom mapping type for your own value objects or classes. For
+example, to map a ``Money`` value object using the `moneyphp/money library`_, you can
+implement a type that converts between this class and a BSON embedded document format.
+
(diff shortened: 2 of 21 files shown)


docs/cookbook/field_type.rst
```diff
@@ -0,0 +1,130 @@
...
+        type_registry: true
+
+Creating a Custom Field Type
+----------------------------
+
+In order to create a custom field type, you need to extend the
+``Doctrine\ODM\MongoDB\Types\Type`` class and implement the required methods.
+
+Mapping a Money Value Object
+-------------------------------------------------
```
greg0ire: This should be shortened

docs/cookbook/field_type.rst
```diff
@@ -0,0 +1,130 @@
...
+    use Doctrine\ODM\MongoDB\Types\ClosureToPHP;
+    use Doctrine\ODM\MongoDB\Types\Type;
+    use InvalidArgumentException;
+    use Money\Money;
+    use Money\Currency;
+
+    #[AsFieldType(Money::class)]
+    final class MoneyType extends Type
+    {
+        // This trait provides a default closureToPHP used during data hydration
```
greg0ire:
```suggestion
        // This trait provides a default closureToPHP() method used during data hydration
```

docs/config.rst
```diff
@@ -317,15 +317,50 @@ your documents.
             </doctrine_mongodb:config>
         </container>
 
-    .. code-block:: php
+You can also use a service as custom type by providing the service id. It's
+necessary to enable the ``type_registry`` by setting.
```
greg0ire: I don't understand this sentence; is it correct?

docs/config.rst
```diff
@@ -317,15 +317,50 @@ your documents.
...
+            <doctrine_mongodb:config
+                type-registry="true"
+            >
+                <doctrine_mongodb:type name="custom_type" service="app.custom_type_service" />
+            </doctrine_mongodb:config>
+        </container>
+
+.. note::
+
+    Defining a custom type as a service requires Doctrine MongoDB ODM 2.16 or higher.
```
greg0ire, line 362: Is it "custom type as a service" or "service as custom type"? I think you should pick one and stick to it, and "custom type as a service" sounds more correct.

src/DependencyInjection/Configuration.php
```diff
@@ -77,6 +81,20 @@ public function getConfigTreeBuilder(): TreeBuilder
                         })
                     ->end()
                 ->end()
+                ->booleanNode('type_registry')
+                    ->defaultFalse()
+                    ->info('If true, create a distinct TypeRegistry for each DocumentManager and inject services tagged with "doctrine_mongodb.field_type". If false, use a shared TypeRegistry for all DocumentManagers.')
```
greg0ire:
```suggestion
                    ->info('If true, create a distinct TypeRegistry for each DocumentManager and inject services tagged with "doctrine_mongodb.field_type" in them. If false, use a shared TypeRegistry for all DocumentManagers.')
```

src/DependencyInjection/DoctrineMongoDBExtension.php
```diff
@@ -467,9 +471,29 @@ public function load(array $configs, ContainerBuilder $container): void
...
-            $configuratorDefinition = $container->getDefinition('doctrine_mongodb.odm.manager_configurator.abstract');
-            $configuratorDefinition->addMethodCall('loadTypes', [$config['types']]);
+        $customTypes = array_map(static function (array $typeConfig): array {
+            if (isset($typeConfig['service'])) {
+                $typeConfig['service'] = new Reference($typeConfig['service']);
+            }
+
+            return $typeConfig;
+        }, $config['types'] ?? []);
+        if (! $config['type_registry']) {
```
greg0ire: Is there a reason to use negative logic here?

src/DependencyInjection/Configuration.php
greg0ire: So when this is set to `false`, there is still a `TypeRegistry`? This is confusing…

src/DependencyInjection/DoctrineMongoDBExtension.php
GromNaN: I started with a flag meaning the opposite, that's why it's negative here and the description is confusing. I'll fix that.

src/DependencyInjection/Configuration.php
GromNaN: There is always a `TypeRegistry` instance, but in one case it's the global shared one (for BC) and when the option is enabled it's 1 instance per document manager.

Do you find a better name for this option?

greg0ire: Maybe `global_type_registry`? Or `shared_type_registry`?

GromNaN: That was my initial idea, but I would like this option to be `false` by default. So, something like `types_per_manager`.

greg0ire: I think you mean `true` by default, and I thought about this too, so I thought about `specific_type_registry` or `per_dm_type_registry`, none of which sound very appealing. If maybe global type registries are going to be deprecated, then I wouldn't sweat it and accept this situation. It's IMO better than an ambiguous setting name.

greg0ire: If I ask Claude, It comes up with:
- `isolated_type_registries`
- `scoped_type_registry`
- `separate_type_registries`
- `type_registry_scope` (enum-style)
- `enable_per_manager_types`
- `type_registry_isolation`

I think only `scoped_type_registry` is appealing. What do you think?

src/DependencyInjection/Configuration.php
```diff
@@ -77,6 +81,20 @@ public function getConfigTreeBuilder(): TreeBuilder
                         })
                     ->end()
                 ->end()
+                ->booleanNode('scoped_type_registry')
```
alcaeus, line 84: Unless we already have a distinct use case for scoped registries, I'd prefer not adding another config setting for this. Using multiple document managers is already dodgy enough, I wouldn't want to add extra complexity.

src/Types/LazyTypeRegistry.php
```diff
@@ -0,0 +1,58 @@
...
+class LazyTypeRegistry extends TypeRegistry
+{
+    /** @param ServiceLocator<Type> $locator */
+    public function __construct(private ServiceProviderInterface $locator)
+    {
+    }
+
+    public function has(string $name): bool
+    {
+        return $this->locator->has($name) || parent::has($name);
```
alcaeus, line 27: I assume there's no significant performance difference between the two (as both are essentially maps connecting a string to an object), so the order of operations does not matter here.

src/Types/LazyTypeRegistry.php
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public function get(string $name): Type
+    {
+        if ($this->locator->has($name)) {
+            return $this->locator->get($name);
+        }
```
alcaeus, line 34: Similar to the comment above, I assume there's no performance issue with checking for a service in the locator or repeatedly fetching the same service from it.

src/DoctrineMongoDBBundle.php
```diff
@@ -77,6 +79,10 @@ public function boot(): void
     private function registerAutoloader(DocumentManager $documentManager): void
     {
         $configuration = $documentManager->getConfiguration();
+        if ($configuration->isLazyGhostObjectEnabled() || $configuration->isNativeLazyObjectEnabled()) {
```
alcaeus, line 82: Should this be an optimisation or bug fix in a separate PR? No strong feelings, it just seems out of place in this PR.

docs/cookbook/field_type.rst
```diff
@@ -0,0 +1,134 @@
...
+    use Doctrine\ODM\MongoDB\Types\ClosureToPHP;
+    use Doctrine\ODM\MongoDB\Types\Type;
+    use InvalidArgumentException;
+    use Money\Money;
+    use Money\Currency;
+
+    #[AsFieldType(Money::class)]
+    final class MoneyType extends Type
+    {
+        // This trait provides a default closureToPHP() used to generate data hydratation classes
```
alcaeus, line 55:
```suggestion
        // This trait provides a default closureToPHP() used to generate data hydration classes
```

src/DependencyInjection/Compiler/TypeRegistryPass.php
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @internal
+ */
+final class TypeRegistryPass implements CompilerPassInterface
+{
+    use PriorityTaggedServiceTrait;
+
+    private const TAG = 'doctrine_mongodb.odm.field_type';
+
+    private const ALL = '\0';
```
alcaeus, line 31: I'll note that this is not actually a null byte, but rather the literal string `\0`. I would hope people don't actually use that.

src/Types/LazyTypeRegistry.php
GromNaN: Indeed, default types are a lot more used than the custom ones. 
```suggestion
        return parent::has($name) || $this->locator->has($name);
```

src/Types/LazyTypeRegistry.php
GromNaN: This is the feature: we can override the default types this way.

src/DependencyInjection/Compiler/TypeRegistryPass.php
GromNaN: Null byte is better, isn't it?
```suggestion
    private const ALL = "\0";
```

But this can be anything that is not a valid document manager name.

src/DependencyInjection/Configuration.php
GromNaN: This is mainly a backward compatibility issue. Currently, people can add types using the static `Type::register()` anywhere is their code.

Without this option. This option will be deprecated in the next major version. We will always have 1 instance of TypeRegistry per DocumentManager.

But, instead I can detect when services are used as type to enable the feature.

src/DependencyInjection/Configuration.php
```diff
@@ -546,10 +564,29 @@ private function addTypesSection(ArrayNodeDefinition $rootNode): void
...
+                                    throw new InvalidArgumentException('Using a service for a MongoDB ODM type requires doctrine/mongodb-odm 2.16 or higher.');
+                                }
+
+                                return $v;
+                            })
                         ->end()
                         ->children()
-                            ->scalarNode('class')->isRequired()->end()
+                            ->scalarNode('class')->end()
+                            ->scalarNode('service')->end()
```
GromNaN, line 589: I will remove this new configuration. As a service needs to be already registered, adding a tag to the service is the way to register it. And the new `#[AsFieldType]` attribute makes it easy.

GromNaN:
I rebased this branch onto the latest 5.6.x and squashed everything into a single commit. I also incorporated the review feedback.

Changes applied:
- Removed the scoped_type_registry configuration option (and the XSD attribute) as discussed with @alcaeus. A TypeRegistry scoped to each document manager is now created automatically whenever a service is used as a field type, detected from the types configuration. Class-based types keep the shared global registry for backward compatibility.
- Replaced the null byte sentinel used in TypeRegistryPass for types shared by all document managers with a readable '*', which cannot collide with a real document manager name.
- Fixed the negative logic flagged by @greg0ire by removing the config switch entirely.
- Reworked the config.rst section to consistently say custom type as a service, removed the stray type-registry attribute from the XML example, and quoted the service shortcut in YAML.
- Shortened the cookbook introduction, fixed the closureToPHP comment (hydratation to hydration), and describe the automatic scoped registry without a configuration option.
- Added the missing tests/Fixtures/Types/CustomTypeService.php fixture that the type registry service test already referenced.

Rebase handling:
- The 5.6.x base had moved with the namespace aliases removal (#971) and the native lazy object changes (#946). I resolved a conflict in DoctrineMongoDBBundle::registerAutoloader() by merging the upstream lazy-object guards with the existing proxy autoloader check, and a conflict in docs/index.rst by keeping both the cookbook/bundles and cookbook/field_type entries.

Validation:
- phpcbf and phpcs are clean on every changed file.
- phpstan reports only two remaining errors, both caused by Doctrine\ODM\MongoDB\Types\TypeRegistry not being present in the released mongodb-odm 2.17.1 installed here. This class comes from the companion mongodb-odm PR (doctrine/mongodb-odm#2966), so the bundle analysis and the type registry tests turn green once that change is available. My changes add no new phpstan errors.
- The config and dependency injection tests pass against the released ODM, including the legacy class-based type path.

The branch now contains a single commit titled Allow custom field types to be injected as services on top of 5.6.x.

GromNaN:
I changed the type registry design following upstream feedback and the companion mongodb-odm work.

Design change:
- Removed the LazyTypeRegistry subclass entirely and deleted src/Types/LazyTypeRegistry.php.
- The per-document-manager registry is now the standard Doctrine\ODM\MongoDB\Types\TypeRegistry, populated through its container constructor: each manager registry receives a ServiceLocator keyed by type name plus the identity map the registry uses to resolve names to services. This mirrors the DoctrineBundle registerInTypeRegistry pattern (doctrine/DoctrineBundle#2221).
- TypeRegistryPass now builds the per-manager service locator from both config-based types (inline class definitions or service references) and attribute-tagged services, with the all-managers plus per-manager object_manager merging kept unchanged.
- Because the ODM TypeRegistry is a concrete, non-subclassed final class, service-based types are lazy-loaded on first access through the locator, and built-in types still fall through.

PR #970, opened 2025-12-16
Mark the definition with the new Document attribute as excluded

GromNaN:
Ensure compatibility with https://github.com/doctrine/mongodb-odm/pull/2933 when the new attribute classes are used.

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -20,3 +20,6 @@ on:
 jobs:
   static-analysis:
     uses: "doctrine/.github/.github/workflows/phpstan.yml@13.1.0"
+    with:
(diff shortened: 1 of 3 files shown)


alcaeus (approved):
LGTM. Feel free to remove the `composer-options` in the static analysis workflow without soliciting an additional review.

.github/workflows/static-analysis.yml
```diff
@@ -20,3 +20,6 @@ on:
 jobs:
   static-analysis:
     uses: "doctrine/.github/.github/workflows/phpstan.yml@13.1.0"
+    with:
+      # Remove once doctrine/mongodb-odm 2.16 is released
+      composer-options: "--minimum-stability=dev"
```
GromNaN:
```suggestion
```

